PR #15, opened 2021-05-01
Make date divider color independent from received message text color

sarbogast:
When the background is light but message bubbles are dark, it's hard to find one text color that works both for received messages and for date dividers. Yet, the same `secondaryTextColor` theme property is used for both. So since the style of date dividers is defined by `subtitle2`, and to stick with the same logic as for `caption`/`captionColor`...

Added a subtitle2Color theme parameter so that the text color of date dividers can be independent from the text color of received message (`secondaryTextColor`). I set the default value of that setting to be the same as `secondaryTextColor` to keep some backward compatibility.

It might make sense to also add one for `subtitle1` but I didn't find any use of this and I wanted to minimize my changes.

### DIFF
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -27,6 +27,7 @@ abstract class ChatTheme {
     required this.sendButtonIcon,
     required this.subtitle1,
     required this.subtitle2,
+    required this.subtitle2Color,
   });
 
   /// Icon for select attachment button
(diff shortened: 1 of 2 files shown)


demchenkoalex (approved):

demchenkoalex:
Hi @sarbogast, thanks for the PR and suggestions! We will revamp theme to include different fonts/colors for different widgets later, to make it more customizable.

demchenkoalex:
Released in [v1.0.2](https://github.com/flyerhq/flutter_chat_ui/releases/tag/v1.0.2) 🙂

PR #18, opened 2021-05-05
Feature/audio message

sarbogast:
This adds the ability to send audio messages in the chat by integrating the flutter_sound package. I tried to make it independent at first, to do what you did for images or attachments, with a few callbacks to let the user implement that part on their own, but since the recording interface has to be deeply integrated with the input bar, it didn't really make sense.

I demonstrated the new feature in the example app, and added some "advanced usage" documentation.

Let me know what you think.


demchenkoalex:
Hi @sarbogast, thank you very much for this! I will take a look as soon as I have time, probably this weekend

sarbogast:
Oh I forgot to mention here that it depends on [another pull request in flutter_chat_types](https://github.com/flyerhq/flutter_chat_types/pull/9). But thanks.

sarbogast:
Also FYI, I plan to add video recordings next

demchenkoalex:
Hi @sarbogast very impressive work! One problem though - `flutter_sound` is licensed under LGPL v3.0 and we cannot use it in out product. Is there any alternatives? By doing a quick search I didn't find anything that would meet our standards (null-safety, web support) and I'm afraid we will need to postpone this until we have some time to write our own plugin (probably as a separate project). But we will for sure take your solution as a base when we get there.

Regarding the video (recording) let's maybe discuss the plan first so we don't end up in this awkward situation?

sarbogast:
Damn! I hadn't considered that. But are you sure you cannot use it? This post seems to suggest it shouldn't be an issue: https://github.com/Canardoux/tau/issues/496

sarbogast:
In any case, I also looked for alternatives, not because of the license but because of the very poor documentation, and I couldn't find anything that worked on the web and had null safety support indeed

sarbogast:
I don't expect video recording to be such an issue though because I don't intend to integrate the recording interface into the chat UI itself

demchenkoalex:
@sarbogast with those kind of licenses I can't be 100% sure, might be tricky in the future, so I would prefer to not use it. Plus, you are saying that the docs are not that good, so maybe writing another good plugin would be a benefit for us and for the community.

Regarding the video I'd like to know what exactly is your plan, cause I'm not sure what video recording means? Like recording a video inside a chat itself and sending it? At the current stage of this project where we lack more basic stuff this might be an overkill to have/support.

sarbogast:
I'm talking about sending recorded video messages, so no live video calling, but just recording a video, uploading it and sharing it in the conversation. I know there might be more pressing issues but the reason why I'm working on these is because I need them for my own project where I'm integrating flutter_chat. The rest of the functionality is mostly enough for what I'm trying to do, I just missed the ability to share audio and video recordings recorded inside the application, not imported from some external recording app.

sarbogast:
Anyway, I need it implemented like I did it here for audio. And for the video, I started doing something similar in the end, taking inspiration from how they do it in Telegram. If you don't want to integrate it like that, it's OK, I just need it done like that because of the way I need the chat to work in my app.

demchenkoalex:
What packages are you planning to use for the video?

sarbogast:
video_player for playback, and camera for recording. Both of them are developed by the Flutter team and support null safety. video_player supports the web, but camera doesn't, which is not an issue for me as my app will be mobile only, but is easy to deactivate when kIsWeb anyway. If you want to have a look, video message with video playback is already implemented [here](https://github.com/sarbogast/flutter_chat_ui/tree/feature/video_message) 
and [here](https://github.com/sarbogast/flutter_chat_types/tree/feature/video_message_types).

demchenkoalex:
Those are much better dependencies indeed :) Will you be able to use your modifications in your project as it is for now? I'd like to merge video messages for sure (not sure about recordings yet) but I still have to do pagination, finish animations and a bunch of other stuff before I can dive into these features.

jinfagang:
@demchenkoalex Hi, will u think video playback feature can be merged for now?

demchenkoalex:
@jinfagang this is audio message, not video. And nothing changed in terms of audio packages, still no good dependencies that I can add. Never worked on a video.

devon:
@sarbogast Could we consider switching to flutter_sound_lite, because flutter_sound makes the app huge.  https://github.com/Canardoux/flutter_sound/issues/291

Yahllil:
I see this feature has been moved to 'Done' but I can't find a way to add Audio Messages in the chat in the documentation. Is this supported?

cc @demchenkoalex

demchenkoalex:
I just closed it cause package changed a lot in the meantime. We will first re-design the code architecture and then come back to this.

Yahllil:
> I just closed it cause package changed a lot in the meantime. We will first re-design the code architecture and then come back to this.

Thanks for your reply @demchenkoalex .
Is this expected to occur in the near future? I am asking because if it won't I'd need to find another way to implement audio messages, but if you guys plan to add it soon I'll wait for your implementation :)

demchenkoalex:
Can't promise anything, since we work when we are eager to implement something. Can happen tomorrow, can happen in a month. Full time job and a family you know :) but it will definitely be there before we think this is a complete solution.

Yahllil:
> Can't promise anything, since we work when we are eager to implement something. Can happen tomorrow, can happen in a month. Full time job and a family you know :) but it will definitely be there before we think this is a complete solution.

Thanks :) Keep up the good work!

marinkobabic:
@sarbogast please see my comment here https://github.com/flyerhq/flutter_chat_ui/issues/189 

Making the target library more abstract by using just an interface we would not have a coupling to the used library

PR #19, opened 2021-05-06
Add authorID as person name in chat

Alexminator99:
I open this pull request with the goal that this can be possible and be optimized. For my purpose it works perfectly, but it needs more work on it. Im noob and this is my first contribution. Thanks for the work with this library.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -163,6 +163,7 @@ class _ChatPageState extends State<ChatPage> {
 
   void _handleSendPressed(types.PartialText message) {
     final textMessage = types.TextMessage(
+      authorId: _user.id,
       id: const Uuid().v4(),
       text: message.text,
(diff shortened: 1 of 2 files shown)


demchenkoalex:
Hi @Alexminator99 thanks for the contribution :) I saw your email, so I will reply here. After we done with a pagination our next priority is an optional sender's name and photo for the message bubble. You can see our board here https://click.flyer.chat/board. So if you can wait a bit please wait, since we will have some design changes coming to support that.

Unfortunately, I can't merge this right now, because you have added a sender's ID to every message which will result in UUIDs being attached to every text message.

Alexminator99:
No worries I will wait. In the meantime those changes work great for me. Awesome 😎 library. I will help you with what I know. Is not much but is something :)

PR #23, opened 2021-05-14
Make preview url loading optional

Elementarereigniss:
The link preview is always on, which can be a security issue for encrypted chats, as the links will be always loaded.
I added the functionality so it can be disabled and added a sentence in the docs.

Also the regex for the link matching matches to bad urls, e.g.  ... or a... or ...x or ._.. etc. (Any use of ellipses or some morse-code will produce a url.) 
Maybe check: https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url

### DIFF
--- a/doc/advanced-usage.md
+++ b/doc/advanced-usage.md
@@ -139,7 +139,7 @@ class _MyHomePageState extends State<MyHomePage> {
 
 ## Link preview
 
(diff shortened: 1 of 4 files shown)


demchenkoalex:
Thanks for the contribution! Also, REGEX and empty paddings were fixed in a preview library.

PR #44, opened 2021-06-14
Text is now selectable

AndreHaueisen:
This is ready to be merged

### What does it do?

`Text` widget is now `SelectableText` for text messages

### Why is it needed?

A user may want to copy a piece of information from the chat to use on another app, for example, a phone number.
`SelectableText` makes that process easier

### How to test it?

Long press the text in a text messages

### Related issues/PRs

Unrelated issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.1.2
+- The text inside text messages is now selectable
+
 ## 1.1.1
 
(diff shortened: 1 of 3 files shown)


demchenkoalex:
Thanks for the contribution! I will merge this now, but in the future we might have a popup on long press with options to reply, delete, copy etc. so this one will be reverted. Or not :)

PR #46, opened 2021-06-14
Made the full composition area trigger input focus (not just the inner field).

muncman:
This is ready to be merged. 

### What does it do?

- Added a `FocusNode` to `_InputState` and its `TextField`. 
- Requesting focus on that node with a `GestureDetector` which wraps the entire composition input area (the part that has the border radius and contains the field itself, etc). 
- The majority of the changed lines are due to increased indentation from the widget wrapping; the only new lines inside `build(...)` are the new, outer `GestureDetector` (and its `onTap`), and `focusNode: _inputFocusNode,` inside the `TextField`.
- Tapping "off" still causes the field to lose focus. 

### Why is it needed?

- This makes it easier for the user to bring up the keyboard, improving usability.
- Before this, if the user tapped a little high or low --but still reasonably "on the input area"-- the chat/app might appear
unresponsive.

### How to test it?

- Tap near the very edge(s) of the input area (where the padding around the `TextField` is), and it should focus the field and bring up the keyboard. 

### Related issues/PRs

- No related issue (I thought I'd just make this PR directly).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
+## 1.1.3
+
+Minor change to make it easier to focus the input field by making the entire composition area tappable.
+
 ## 1.1.2
+
 - The text inside text messages is now selectable
 
 ## 1.1.1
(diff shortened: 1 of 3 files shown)


demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #51, opened 2021-06-18
Adds the possibility to customize the date format

AndreHaueisen:
### What does it do?

- I've created an optional property called `customDateFormat` that allows the user to have custom dates
- I've also done a very small code optimization where an error in `ChatList` will return `const SizedBox` instead of `Container`

### Why is it needed?

We would have more flexibility when showing dates

### How to test it?

`Chat` now has a property called `customDateFormat`. To test, pass a `DateFormat` to it

### Related issues/PRs

None

Simulator Screen Shot - iPhone 12 Pro Max - 2021-06-18 at 12 40 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.5
+
+- Adds [DateFormat] `customDateFormat` allowing the user to customize the date
+
 ## 1.1.4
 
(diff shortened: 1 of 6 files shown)


demchenkoalex:
Thanks for the PR!

PR #53, opened 2021-06-22
On text changed callback

AndreHaueisen:
### What does it do?

Adds a callback so we can know when the text changed

### Why is it needed?

It is useful when we want, for example, show a typing indicator. In that case, we need a way to know when the user started typing.

### How to test it?

Pass a `onTextChanged` callback to `Chat`

### Related issues/PRs

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.5
+
+- Adds [onTextChanged] callback to detect when the user is typing
+
 ## 1.1.4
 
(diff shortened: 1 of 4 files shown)


demchenkoalex:
Thanks for the PR!

jinfagang:
this params no longer existed?

demchenkoalex:
inside `InputOptions`. You can also search the codebase, it is open source :) also mentioned in the `CHANGELOG`

PR #57, opened 2021-06-29
Custom empty state

AndreHaueisen:
### What does it do?

It allows you to customize what the user sees when there are no messages

### Why is it needed?

A string might not be enough in certain cases. You might want to retry the request in case you get an error.
A custom widget could come in handy to add, for example, a button

### How to test it?

Give `emptyState` a `Widget`

### Related issues/PRs

No


demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

Issue #65, opened 2021-08-04
AvatarImage doesn't always cover userAvatarNameColor circle


m-j-g:
This is probably much more noticeable on dark themes with bright userAvatarNameColors, but a lot of images, even ones that are perfect squares, aren't covering 100% of the avatar color circle behind it.  This leads to an extremely thin line of color that isn't smooth and kinda makes it look bad.  Attached a picture.  I have the avatar color set to a green in this one.

![IMG_624B5B46A0B2-1](https://user-images.githubusercontent.com/10711395/128252401-acba6513-3848-4d76-b656-efa52cc8dee0.jpeg)

demchenkoalex:
Hi, thanks for reporting, will take a look

demchenkoalex:
Released in [v1.2.0](https://github.com/flyerhq/flutter_chat_ui/releases/tag/v1.2.0).

PR #77, opened 2021-08-26
Feature/add background color for imaged avatar

pierrebarbaroux:
Hello! thanks for this great plugin that has been very useful for me 🙌 

I encountered some limitations for a specific case, which is described below, and thought it might be useful to publish my contribution to resolve it, in order to make the library more flexible. 

### What does it do?

- add parameter `userImagedAvatarBackgroundColor` to `ChatTheme` ;
- use it as user avatar `CircleAvatar`'s background color, in the case an image is used.

---

### Why is it needed?

A user avatar with an image and **transparent parts** has currently a green background color which is not customizable.
This contribution allows people to customize this color.

#### The current default behavior (<= [v1.2.0 release](https://github.com/flyerhq/flutter_chat_ui/releases/tag/v1.2.0))
old

#### The new default behavior, for the `DefaultChatTheme`
new-light

#### The new default behavior, for the `DarkChatTheme`
new-dark

The existing parameter `userAvatarNameColors`, which is used for username color and avatar background color behind user initials if no image is provided, could not really be used here. This new color is mostly driven in an esthetic way, in order to fit with the other `ChatTheme` parameters.

---

### How to test it?

Put a transparent image as a user avatar (displayed on received messages only) and you should not see the green default background color. Customize it by defining a new `userImagedAvatarBackgroundColor` color to your theme.

---

### Related issues/PRs

None

### DIFF
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -79,6 +79,7 @@ abstract class ChatTheme {
     required this.sentMessageLinkTitleTextStyle,
     required this.userAvatarNameColors,
     required this.userAvatarTextStyle,
+    required this.userImagedAvatarBackgroundColor,
     required this.userNameTextStyle,
   });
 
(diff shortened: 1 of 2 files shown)


demchenkoalex:
Hello Pierre, thanks for your contribution! I have one question - we had this issue #65, and previously background colour was the same as the colour of the name inside the bubble. We had to set it to `null` because that issue. If I will merge this we will have the same situation and users will need to specifically pass transparent colour.

Additionally, by design, avatar colours match user names inside bubbles and colours selected are based on message author's `id` hash, these colours are customisable. Adding another colour to customise customisable colours doesn't seem reasonable to me.

Do we have another options?

PR #145, opened 2021-11-04
Add Traditional Chinese L10n

Wei-Hsun:
* add Traditional Chinese L10n

### What does it do?

It adds support for Traditional Chinese (繁體中文-台灣).

### Why is it needed?

It can helps Taiwan user to use this project.

### How to test it?

Provide "const ChatL10nZhTW()" for property "l10n".

### Related issues/PRs

None.


demchenkoalex:
Thank you for the contribution!

PR #181, opened 2021-12-24
feature to tag users in group chat

Mufaddal1125:
### What does it do?

was looking for a way to popup a list of users to tag when you type '@' like WhatsApp. this PR implements optional parameter `mentions` which accepts list of `Mention` objects to add mentions data

### Why is it needed?

Needed a feature to tag users in a group chat

### How to test it?

add `mentions` parameter in `Chat` widget add `Mention(trigger: '@', data: [{'id': '1', 'display': 'John Doe'}])` to mentions, type `@` in input and you'll see list of users provided in data in mention

### Related issues/PRs
Not related to any issues/PRs


demchenkoalex:
Hey @Mufaddal1125 thanks for the contribution! You can always ask before doing an MR, so you don't end up wasting your time. From what I see this code depends on `https://github.com/rrousselGit/flutter_portal` which is currently not actively supported and is looking for maintainers. I can't add dependencies that will not be updated shortly for Flutter 3.0 for example (and every major release), so I will postpone this MR and see if dependencies used here will receive some updates. Thanks for the understanding.

Mufaddal1125:
Oh, didn't knew that the dependency is not maintained anymore. anyways thanks for the review @demchenkoalex.

demchenkoalex:
Hey @Mufaddal1125 I see that you're tailoring this to your needs. Can you fork the lib instead? cause now it can't be merged in any case.

m-j-g:
Bummer, I get requests to be able to @ people all of the time.

Mufaddal1125:
Oops! didn't realize those commits would be added to the PR, I've now removed them, sorry for the inconvenience caused.

felixgabler:
This could really benefit from using `flutter_parsed_text`. Once the user has selected their mention, we could write it into the text message as `[@<USER_ID>]` and then replace it with the latest name and a link-type style by using the library when the message is displayed.

demchenkoalex:
I guess so, just would need to be re-written using `flutter_parsed_text`, since we have it already

PR #203, opened 2022-01-30
localize: add German translations

felixgabler:
### What does it do?

Allow users to show German translations

### Why is it needed?

Translations are always useful ;)

### How to test it?

Specify `ChatL10nDe` as `l10n` argument of `Chat` widget.

### Related issues/PRs

N/A


demchenkoalex:
Thanks for the PR!

PR #205, opened 2022-01-30
feat: allow overriding of link style

felixgabler:
### What does it do?

Specifying a `sentMessageBodyLinkTextStyle` or `receivedMessageBodyLinkTextStyle` in the `ChatTheme` will, optionally, override the text style used for the link preview of text messages.

### Why is it needed?

I faced an issue where I set a text style for body text in the chat theme but this style has a different decoration setting such that the underline also gets removed. Now, there is more flexibility.

### How to test it?

### Related issues/PRs


demchenkoalex:
Thanks for the PR!

PR #206, opened 2022-01-30
feat: add incomplete markdown support for text messages

### LINKED ISSUE #157: Possible markdown support?
### Discussed in https://github.com/flyerhq/flutter_chat_ui/discussions/154

<div type='discussions-op-text'>

<sup>Originally posted by **luohao123** November 10, 2021</sup>
Just wonder if there any option to make users enable markdown support.

I noticed that there already have link support, but bold and ordered check list does't support</div>

felixgabler:
### What does it do?

Add simple markdown rendering and replace link renderer with it too. Closes #157
 
### Why is it needed?

Most modern chat apps have markdown support. Also, it is a bit cumbersome to maintain the header + text rendering logic here and in the link previewer package as well. Now, it is unified across both.

### How to test it?

I added a simple Markdown example in the example app with link and email as well.

### Open Questions / TODO

- Should there be an option to disable Markdown formatting in the `Chat` constructor?

### DIFF
--- a/example/android/app/src/main/AndroidManifest.xml
+++ b/example/android/app/src/main/AndroidManifest.xml
@@ -31,4 +31,18 @@
             android:name="flutterEmbedding"
             android:value="2" />
     </application>
+
+    <!-- Required for url and mailto launching via url_launcher -->
+    <queries>
+        <!-- If your app opens https URLs -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="https" />
+        </intent>
+        <!-- If your app sends emails -->
+        <intent>
+            <action android:name="android.intent.action.SEND" />
+            <data android:mimeType="*/*" />
+        </intent>
+    </queries>
 </manifest>
--- a/example/assets/messages.json
+++ b/example/assets/messages.json
@@ -8,6 +8,18 @@
     "createdAt": 1598438797000,
     "id": "e7a673e9-86eb-4572-936f-2882b0183cdc",
     "status": "seen",
+    "text": "## Heading\n\n**This is bold text**\n\n_This is italic text_\n\n~~Strikethrough~~\n\nInline `code`\n\n1. Lorem ipsum dolor sit amet\n2. Consectetur adipiscing elit\n3. Integer molestie lorem at massa\n\n+ Create a list by starting a line with `+`, `-`, or `*`\n+ Sub-lists are made by indenting 2 spaces:\n  - Marker character change forces new list start:\n    * Ac tristique libero volutpat at\n    + Facilisis in pretium nisl aliquet\n    - Nulla volutpat aliquam velit\n+ Very easy!",
+    "type": "text"
+  },
+  {
+    "author": {
+      "firstName": "John",
+      "id": "b4878b96-efbc-479a-8291-474ef323dec7",
+      "imageUrl": "https://avatars.githubusercontent.com/u/14123304?v=4"
+    },
+    "createdAt": 1598438797000,
+    "id": "e7a673e9-86eb-4572-936f-2882b0183cdc",
+    "status": "seen",
     "text": "https://flyer.chat",
     "type": "text"
   },
--- a/example/assets/messages.json
+++ b/example/assets/messages.json
@@ -8,7 +8,7 @@
     "createdAt": 1598438797000,
     "id": "e7a673e9-86eb-4572-936f-2882b0183cdc",
     "status": "seen",
-    "text": "**bold** _italic_ ~strikethrough~ `code`",
+    "text": "**bold** _italic_ ~strikethrough~ `code`\nhttps://link-test.com test@email.com",
     "type": "text"
   },
   {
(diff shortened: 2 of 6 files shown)


demchenkoalex:
I will discuss your questions with our designer, thanks for this :)

demchenkoalex:
had a brief chat - we should only support like basic features - **bold**, _italic_, ~~strikethrough~~, `maybe code`. No need to support tables and headers really yes?

felixgabler:
Sure, sounds reasonable. I'll submit a new PR then once I get to it

felixgabler:
I changed my PR to use `flutter_parsed_text` instead of `flutter_markdown`. Now, it will display **bold**, _italic_ and ~strikethrough~ text. I can also add code if you agree with my implementation. Another cool thing is that we could build @mentions this way too. I can follow up with a PR for that once this one is ready to merge.

One more thing I would like to propose is to slightly rearchitecture the link_previewer package. Wouldn't it make more sense to pass a content widget so that we can reuse the markdown parser and keep the text the same which is displayed above the link preview. This would mean that link_previewer is only responsible for the link preview and not for rendering links anymore. I could easily add link detection and opening to the `ParsedText`. We could remove the dependency on `flutter_linkify` and responsibilities for text rendering would not have to be shared anymore.

demchenkoalex:
Well, link preview is a bit separate project, and links have titles and descriptions, this way if you use link preview in your app you will not get those. So it must stay as it is, we can only modify it for the chat I guess, but then I don't really know what will happen with all the animations and stuff, because link preview will only return image and how it will be appended to the text message.. not sure.

felixgabler:
What I am talking about is just adding a field to the link_preview package where one can supply a Widget that gets rendered instead of the plain text. The link preview data title, description and image would stay as they currently are. Would that be okay?

demchenkoalex:
Ah yes, that's fine. Can you do a PR there? Or I can do it later

felixgabler:
Here is the PR: https://github.com/flyerhq/flutter_link_previewer/pull/38

demchenkoalex:
Hi @felixgabler merged the link previewer thing, you can do your magic now :) Thank you!

felixgabler:
Awesome! What do you think we should do in terms of parametrization? We currently have linkify options but that would perhaps not be relevant anymore.. any options you'd like me to include?

demchenkoalex:
What you already have is enough I think, everything else will be added on request and if feasible. You are free to include whatever you think is necessary.

felixgabler:
One thing that will have to be added to the documentation is that for url opening and mailto functionality, we need to configure the `url_launcher` package correctly. This is described here: https://pub.dev/packages/url_launcher and should also be added to the link previewer package documentation.

felixgabler:
Once this is merged, I can take a look at the @mention feature

felixgabler:
Hi, I don't want to hassle you but was just wondering when you think you might have to look at the PRs? Best!

demchenkoalex:
Hey, sorry, I am from Ukraine and I didn't expect a war to happen, so open-source went to the second plan. Now I am away from computer and will come back in April, I want to resolve all pending things then. Thanks for understanding.

demchenkoalex (approved):

demchenkoalex:
Thank you very much for the PR!

jinfagang:
@demchenkoalex Looks promising! But will it break all current users using flychatui? (Personally speakking I really like the simplicity flyerui API desgin in dart. its easy and modualized. If new work can compatible with API that would be pretty nice!)

demchenkoalex:
@jinfagang it will be rather v2 breaking kind of change. Simplicity will be replaced with modularity, but I will try to provide a good defaults, so plug and play is still possible

mingxin-yang:
@demchenkoalex  I am also trying to use Markdown to display the code. I want to change the ParsedText used in TextMessageText through textMessageBuilder and replace it with https://pub.dev/packages/flutter_markdown, but I keep making mistakes. I must fix the height in the Container, otherwise an error will be reported, no know what's wrong
```
// textMessageBuilder: (
        //   types.TextMessage message, {
        //   required int messageWidth,
        //   required bool showName,
        // }) {
        //   return Container(
        //     height: 100,
        //     width: messageWidth.toDouble(),
        //     child: Markdown(
        //       selectable: true,
        //       data: message.text,
        //       styleSheet: MarkdownStyleSheet(
        //         codeblockDecoration: BoxDecoration(
        //           color: Theme.of(context).backgroundColor,
        //           borderRadius: BorderRadius.circular(4.0),
        //         ),
        //         codeblockPadding: EdgeInsets.all(8.0),
        //         code: TextStyle(
        //           fontFamily: 'SpaceMono',
        //           color: Theme.of(context).primaryColor,
        //           fontSize: 16,
        //           backgroundColor: Colors.transparent,
        //         ),
        //         p: TextStyle(
        //           fontSize: 16,
        //         ),
        //       ),
        //       extensionSet: md.ExtensionSet(
        //         md.ExtensionSet.gitHubFlavored.blockSyntaxes,
        //         [
        //           md.EmojiSyntax(),
        //           ...md.ExtensionSet.gitHubFlavored.inlineSyntaxes
        //         ],
        //       ),
        //     )
        //   );
        // },
```

PR #207, opened 2022-01-31
Allow author override through `resolveAuthor`

felixgabler:
### What does it do?

Add `resolveAuthor(String id)` method to the `Chat()` which can be used to fetch the latest user information to display in avatar and message name headers.

### Why is it needed?

Because the author currently comes from the message data, changes of profile picture or name only affect new messages (or all old messages would need to be replaced). It should be possible to optionally override user information on the top level.

### How to test it?

Add `showUserNames` and `showUserAvatars` to the Chat and specify a `resolveAuthor` function that returns newer information.

### Related issues/PRs

The other option would be specifying `nameBuilder` and `avatarBuilder` which has a wider scope and effects than this PR. If you think it would be more useful, I can also create a PR with that functionality. However, this is the smallest change necessary to solve my needs.

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -177,8 +177,7 @@ List<Object> calculateChatMessages(
       'nextMessageInGroup': nextMessageInGroup,
       'showName': notMyMessage &&
           showUserNames &&
-          showName &&
-          getUserName(message.author).isNotEmpty,
+          showName,
       'showStatus': true,
     });
 
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -175,9 +175,7 @@ List<Object> calculateChatMessages(
     chatMessages.insert(0, {
       'message': message,
       'nextMessageInGroup': nextMessageInGroup,
-      'showName': notMyMessage &&
-          showUserNames &&
-          showName,
+      'showName': notMyMessage && showUserNames && showName,
       'showStatus': true,
     });
 
--- a/lib/src/widgets/text_message.dart
+++ b/lib/src/widgets/text_message.dart
@@ -12,6 +12,7 @@ class TextMessage extends StatelessWidget {
   /// Creates a text message widget from a [types.TextMessage] class
   const TextMessage({
     Key? key,
+    required this.author,
     required this.emojiEnlargementBehavior,
     required this.hideBackgroundOnEmojiMessages,
     required this.message,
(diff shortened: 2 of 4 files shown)


demchenkoalex:
Still don't get how this works? So this function is called on every message? How's author will be resolved for each different message?

felixgabler:
In our use case, we have a map of all the members of a group (`Map<String, User>`). This map contains newer information than what is stored in each message and it can be accessed by the ID like this:

```
Chat(
   ...
   resolveAuthor: (id) => _userMapById[I'd],
)
```

felixgabler:
One implication of this is if you e.g. have a listener on all the group members data that updates the name/pictures, we would not have to reload all messages and replace their author fields. Instead, only the currently visible messages would get rebuilt.

demchenkoalex:
hey @felixgabler don't know how to contact you otherwise, I have quite a few things to finish at work atm, so I didn't have time to finish all the PRs. I should come back this week/weekend. Thanks!

lib/src/util.dart
```diff
@@ -175,10 +175,7 @@ List<Object> calculateChatMessages(
     chatMessages.insert(0, {
       'message': message,
       'nextMessageInGroup': nextMessageInGroup,
-      'showName': notMyMessage &&
-          showUserNames &&
-          showName &&
-          getUserName(message.author).isNotEmpty,
+      'showName': notMyMessage && showUserNames && showName,
```
demchenkoalex, line 178: We remove `getUserName(message.author).isNotEmpty` from here, it is used in 2 places in `text_message`, you covered only one. My proposition here is instead of doing it in `text_message` like you did `if (showName && name.isNotEmpty)` we do it in `message.dart`: look for `case types.MessageType.text:` you see how `showName` is passed to both `textMessageBuilder` and `TextMessage`. Instead of doing `showName: showName,` in both cases we do `showName: showName && name.isNotEmpty` where name you could get from `author` parameter you already added.

lib/src/widgets/text_message.dart
```diff
@@ -12,6 +12,7 @@ class TextMessage extends StatelessWidget {
   /// Creates a text message widget from a [types.TextMessage] class
   const TextMessage({
     Key? key,
+    required this.author,
```
demchenkoalex, line 15: if this was needed to replace all occurrences of `message.author.id` or `message.author` - please check through the project, it is used in file, image messages and in some other places - they should be covered too? or this is added only here just because only here we can show the name of the user and in other places we are interested in id?

demchenkoalex:
Hey @felixgabler could you take a look at the questions here? I am ready to merge, but those 2 things have valid points I think

Issue #20, opened 2021-05-12
feature: user avatar


m-j-g:
Hey guys. Just discovered your package and appreciate how professional all your material looks.  
Is it possible for us to show a user avatar next to the chat bubble, or how do you tell people apart in group chats?  Also, do you have any interest in “reaction” so that people can “like” a comment?

demchenkoalex:
Hi @m-j-g right now it is not possible to show the avatar, but we will add it soon, we already have design and everything ready for that. It is on our board https://click.flyer.chat/board. Reactions are planned, but later. We will consider prioritizing it if maybe more people will comment here, because we lack a lot of chat features and need to choose what to do next :)

m-j-g:
> Hi @m-j-g right now it is not possible to show the avatar, but we will add it soon, we already have design and everything ready for that. It is on our board https://click.flyer.chat/board. Reactions are planned, but later. We will consider prioritizing it if maybe more people will comment here, because we lack a lot of chat features and need to choose what to do next :)

Cool, thanks for the response.

felixgabler:
This would really be great for us too!

Ajay1901:
Hey, your library is awesome and helping in a great way for our app. But we feel you are lacking a very vital feature... i.e, in groupRooms the message author's name is not there (who sent what). Are you looking forward to implement anytime soon? If yes any sort of ETA you have in mind?

demchenkoalex:
Hi @Ajay1901 yes, I am looking forward to implement it. As you can see in the pagination issue I've been stuck with Flutter animations for over a month now, finding a lot of strange behaviors and issues. This pushed back features delivery a lot :( I work only weekends and I can't tell you it will be delivered on XX of May, but it will be the first thing to do after I finish replacing ListView with SliverAnimatedList and adding pagination to it. I can't also jump on this feature because SliverAnimatedList affected the design and code structure a lot so I better finish it first. Then this, and this feature will probably take up to 1 day because it will be easy to add.

Ajay1901:
Hey, thanks a lot to reply. I wanted to ask if you can guide me roughly on where to make the changes, I'll give it a shot and raise a PR? Also, there's no direct way of sharing videos I think we have types of the message as File, Images, and text.  Do I need to consider video as a part of File only...!

demchenkoalex:
[...]

PR #222, opened 2022-02-19
Add dateHeaderBuilder

arsamme:
### What does it do?
I've added `dateHeaderBuilder` to add ability to customize date header widget.

### Why is it needed?

For customizing date header widget. For example I wanna add shadow, border or etc. to date header.

### How to test it?

If you provide `dateHeaderBuilder` it should get called to build date header and return your custom widget. Else, default date header widget will be returned.


arsamme:
I've used `showStatus` from `message` so this PR depends on [#20](https://github.com/flyerhq/flutter_chat_types/pull/20)

demchenkoalex:
Thanks for the PR!

PR #223, opened 2022-02-21
Add options to open preview on title or image tap

felixgabler:
### What does it do?

You can now specify options to enable the tap-to-open behavior from yesterdays PR.

### Why is it needed?

No other way of configuring this yet.

### How to test it?

### Related issues/PRs

[Let us know if this is related to any issue/pull request.](https://github.com/flyerhq/flutter_link_previewer/pull/39)

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -60,8 +60,7 @@ class Chat extends StatefulWidget {
     required this.onSendPressed,
     this.onTextChanged,
     this.onTextFieldTap,
-    this.openOnPreviewImageTap = false,
-    this.openOnPreviewTitleTap = false,
+    this.previewTapOptions = const PreviewTapOptions(),
     this.scrollPhysics,
     this.sendButtonVisibilityMode = SendButtonVisibilityMode.editing,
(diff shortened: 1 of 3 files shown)


demchenkoalex:
Not that familiar with Flutter, but is there a way to pass like all LinkPreview properties somehow? In React Native I would pass props: LinkPreviewProps and then just do <LinkPreview {...props} /> and user can configure everything.. here we adding two props, but still some are not configurable, and if we add more props, we need to copy them over and over.. don't like Dart because of that :D

felixgabler:
Unfortunately, that is not possible. We could however create a configuration object that gets passed down to simplify all future configuration. Do you want me to do that here?

demchenkoalex:
Yes, please. I will see how it is done and will try to apply that pattern in the future.

felixgabler:
I just pulled them into one config object which people sometimes do so that they don't have to change many files when adding or removing. Basically, like context. I wish there was something like `...props`. Makes me miss React :(

demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #231, opened 2022-03-06
allow using custom avatar builder to always fetch newest profile

dariuspo:
What does it do?
Adding avatar builder that can be future builder which can be used to fetch the latest user information to display in avatar.

Why is it needed?
Because the author currently comes from the message data, changes of profile picture only affect new messages (or all old messages would need to be replaced). It should be possible to optionally override Avatar information on the top level. 

How to test it?
add param avatarBuilder that accept Widget Function(String userId), with this we can access the user id for that message and use our own widget to fetch user avatar

Related issues/PRs
https://github.com/flyerhq/flutter_chat_ui/pull/207
This PR not includes user name builder, because i believe we can already override using existing builder.

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -67,7 +67,7 @@ class Chat extends StatefulWidget {
     this.textMessageBuilder,
     this.theme = const DefaultChatTheme(),
     this.timeFormat,
-    this.userBuilder,
+    this.avatarBuilder,
     this.usePreviewData = true,
     required this.user,
   }) : super(key: key);
@@ -213,7 +213,7 @@ class Chat extends StatefulWidget {
 
   /// This is to allow custom user builder
   /// By using this we can fetch newest user info based on id
-  final Widget Function(String userId)? userBuilder;
+  final Widget Function(String userId)? avatarBuilder;
 
   /// Show user names for received messages. Useful for a group chat. Will be
   /// shown only on text messages.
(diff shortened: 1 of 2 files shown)


demchenkoalex:
Could this be combined with #207 ? I don't want to have some bits configurable through some functions here, and some bits there, better to have a unified solution for messages data, not just avatar.

felixgabler:
This is one part of the alternative I mentioned in #207. We would need both a `avatarBuilder` and a `nameBuilder` to allow overriding the author completely. I could see this PR here becoming more useful if a builder for name is added and we also extract the default avatar and name building into their own widgets.

Then, it would be easy to override author properties while still allowing users to use the design provided by this library and not introducing additional maintenance effort.

Here is how I could see this working (the `_resolveUser` method is what I would have passed into the chat in #207):
```dart
Chat(
   ...
   avatarBuilder: (id) => ChatAvatar(imageUrl: _resolveUser(id).picture),
   nameBuilder: (id) => ChatNameHeader(firstName: _resolveUser(id).firstName, lastName: _resolveUser(id).lastName),
);

my.User _resolveUser(String id) => _newUserInfoMap[id];
``` 

On a side note, extracting widgets is something that could be used more inside the project anyways since it is the Flutter way and preferred over many `_buildXYZ` methods.

demchenkoalex:
Will you be able to combine these 2 @felixgabler ? also I mentioned 2 things on #207 - will happily merge then :)

felixgabler:
Sure, I could build on top of this PR here but also do not want to intrude into what @dariuspo had planned. I would definitely close #207 though if we build the functionality I just described.

demchenkoalex:
Thanks for the PR and the idea @dariuspo!

Issue #216, opened 2022-02-07
Emoji only message horizontal margin is set to zero


munkius:
Hi!

Great job on the library. I was wondering why the horizontal margin is set to `0.0` for displaying an emoji-only message, and having `EmojiEnlargementBehavior.multi` enabled (by default). This makes the horizontal space skewed as opposed to giving it the room I'd expect, using the theme's configured `messageInsetsHorizontal`.

#### Current behavior
![image](https://user-images.githubusercontent.com/194593/152786248-944a1317-4bcb-40a2-8f73-f91235e683da.png)

#### Expected behavior
![image](https://user-images.githubusercontent.com/194593/152786097-fbd3061a-f352-4f1e-a6ec-0299e35085ed.png)

#### The line of code where it happens
https://github.com/flyerhq/flutter_chat_ui/blob/d299c5551c2f6e4118a8d2976a8050317e400a8e/lib/src/widgets/text_message.dart#L158

munkius:
Anyone? :)

demchenkoalex:
You can always open a PR if you want to speed things up. I saw it and as soon as I will be done with PRs I will check.

munkius:
Voila!

demchenkoalex:
Will be included in a next release - `1.5.6`. Thank you!

munkius:
You're so welcome! Keep up the good work 💪

PR #233, opened 2022-03-08
Respecting margin for emoji-only messages

munkius:
### What does it do?

It's a minor change, regarding the horizontal margin in text messages.

### Why is it needed?

Text messages containing just 1 or more emoji's did not respect any horizontal margin set. It's always defaulted to `0.0`. Hence these kind of messages look different from all the other messages. 

### How to test it?

Create a message with just one or more emoji characters.

### Related issues/PRs

It fixes issue #216.


demchenkoalex:
Thanks for the PR!

PR #236, opened 2022-03-11
feat: increase send button tap size

felixgabler:
### What does it do?

Increase the tap target size of the send icon. For this, I had to apply the padding a bit differently and convert EdgeInsetsGeomtry to EdgeInset for this (which is anyways used most of the times).

Also, keep in mind that I extended the padding on the left of the send button by 12pt. Should we reduce the margin on the left side of the send button by that amount? It would be weird to not allow tapping to the left of the button.

### Why is it needed?

Before, the send icon could only be tapped in a 24pt box. This can sometimes be rather fiddley and annoying, to some degree.

### How to test it?

### Related issues/PRs

### DIFF
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -104,7 +104,7 @@ abstract class ChatTheme {
   final Color backgroundColor;
 
   /// Margin around date dividers
-  final EdgeInsetsGeometry dateDividerMargin;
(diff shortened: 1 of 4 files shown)


demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #245, opened 2022-04-10
feat: add scroll controller as a parameter

Faaatman:
### What does it do?

Scrollcontroller can now be passed to the Chat.
this would add the flexibility of controlling the chat scroll view outside the widget

### Why is it needed?

To be able to control the scroll view. ex: a widget to scroll to the bottom of the chat.
I have a widget that is shown whenever a new message is received and the user is scrolled up in the chat, upon tapping the widget it scrolls to the end of the list.

### How to test it?

Create a `scrollController` and pass it to the chat widget.

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -73,6 +73,7 @@ class Chat extends StatefulWidget {
     this.timeFormat,
     this.usePreviewData = true,
     required this.user,
+    this.scrollController,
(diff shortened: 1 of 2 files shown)


Faaatman:
This can be used as an alternative until #209 is done.

demchenkoalex:
Thanks for the PR!

Faaatman:
Thank you for your quick response🙏🏼

PR #246, opened 2022-04-10
Make attachment button margin configurable

felixgabler:
### What does it do?

Add option to theme that sets the margin for the attachment button analogously to how it is done for the send button.

### Why is it needed?

More styling control

### How to test it?

### Related issues/PRs


demchenkoalex:
Thanks for the PR!

PR #247, opened 2022-04-10
Allow showing loading spinner around file icon

felixgabler:
### What does it do?

Let users specify which file message to show with a loading spinner.

Screenshot 2022-04-10 at 13 32 30

### Why is it needed?

When users store files online, they will have to download them before opening. This can take time and the UI should reflect it.

### How to test it?

### Question

I implemented this through a top level field (`fileMessageLoadingId`) in chat. The alternative would have been to set it inside the `FileMessage` but that would have meant a change in the types repository and also could be overridden much easier (if new messages come in, they will be parsed without the loading and it would get lost).

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -41,7 +41,6 @@ class Chat extends StatefulWidget {
     this.emojiEnlargementBehavior = EmojiEnlargementBehavior.multi,
     this.emptyState,
     this.fileMessageBuilder,
-    this.fileMessageLoadingId,
     this.groupMessagesThreshold = 60000,
     this.hideBackgroundOnEmojiMessages = true,
     this.imageMessageBuilder,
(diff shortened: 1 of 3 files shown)


demchenkoalex:
Maybe we should strive for more reusable solution? Now you can show spinner only on one message at the time, plus ids are being thrown in the UI library. Maybe instead a file message (in the types package) should have loading status and copyWith function to change that?

felixgabler:
> Maybe we should strive for more reusable solution? Now you can show spinner only on one message at the time, plus ids are being thrown in the UI library. Maybe instead a file message (in the types package) should have loading status and copyWith function to change that?

That's the other solution I talked about above. This would mean that when new messages come in, they might override the loading status because all messages are parsed again. However, one could also say that the responsibility to keep it persistent is on the user's side. So yes, I can change it to allow specifying it on the file message level.

demchenkoalex:
loading will be stored in the messages array, so I don't see how it will be automatically overridden, before you call `copyWith` specifically

felixgabler:
I have just created a PR in the types package. How would you like me to handle the dependencies? Wait until `flutter_chat_types` has another release (probably better) to use that or switch to pulling it from git for a while?

demchenkoalex:
Will check and merge/release now

felixgabler:
Done too :P

demchenkoalex:
Will check later, need to work now :)

demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #248, opened 2022-04-10
Extract reusable widgets for name and avatar and add `avatarBuilder` and `nameBuilder`

felixgabler:
### What does it do?

As described in #231, it is now possible to both override the name and avatar builder but also to reuse the current design with updated author data.

### Why is it needed?

For one, we do not want to have to update the author fields for all messages when the profile picture changes.

Also, it should be possible to specify how the avatar is rendered. This allows us to add things like a loading spinner and such, or reuse already existing image caching logic.

### How to test it?

### Related issues/PRs

Builds on top of #231, much of the credit goes to @dariuspo. Will make #207 irrelevant.


demchenkoalex:
Thanks for the PR!

PR #254, opened 2022-04-15
Swipe to reply prototype

Voidozzer:
### What does it do?

Adds swipe to reply prototype

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -80,7 +80,7 @@ SPEC CHECKSUMS:
   open_file: 02eb5cb6b21264bd3a696876f5afbfb7ca4f4b7d
   SDWebImage: 0905f1b7760fc8ac4198cae0036600d67478751e
   SwiftyGif: 6c3eafd0ce693cad58bb63d2b2fb9bacb8552780
-  url_launcher_ios: 02f1989d4e14e998335b02b67a7590fa34f971af
+  url_launcher_ios: 839c58cdb4279282219f5e248c3321761ff3c4de
 
 PODFILE CHECKSUM: aafe91acc616949ddb318b77800a7f51bffa2a4c
 
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -80,7 +80,7 @@ SPEC CHECKSUMS:
   open_file: 02eb5cb6b21264bd3a696876f5afbfb7ca4f4b7d
   SDWebImage: 0905f1b7760fc8ac4198cae0036600d67478751e
   SwiftyGif: 6c3eafd0ce693cad58bb63d2b2fb9bacb8552780
-  url_launcher_ios: 839c58cdb4279282219f5e248c3321761ff3c4de
+  url_launcher_ios: 02f1989d4e14e998335b02b67a7590fa34f971af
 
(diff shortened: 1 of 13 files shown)


Justus-M:
Very excited for this! Thank you!

demchenkoalex:
Hi @felixgabler! We are a bit stuck here and maybe you will have some ideas. We are using https://pub.dev/packages/flutter_slidable for the swipe to reply feature, but the package has one problem - it has no auto close, i.e. when we swipe to reply we want the reply icon to hide automatically, this package does not support it. I guess it was mentioned somewhere there, but it is not actively maintained. @Voidozzer wrapped slidable with Listener (see file `message.dart`) but it detects both vertical and horizontal swipes resulting in reply triggering while you are simply scrolling through messages (you can observe this behaviour on a branch). Maybe you can take a look with a fresh pair of 👀

felixgabler:
Hi! Sure, I will have a look tomorrow when my eyes are actually fresh again 😴 I'll need to check this PR out to see where the problem is exactly.

Until then, one quick thing I found is this package: https://pub.dev/packages/swipe_to. The code looks nice and short (I wouldn't expect this functionality to require a ton of code). It seems to have swipe detection and showing an icon if that is something we need.

Otherwise, one could fork the repository of flutter_slidable but that obviously always brings extra maintenance.

demchenkoalex:
Thanks! We could totally try a different dependency if needed. Fork would not help because when distributing a package you can have only pub dependencies, we would need to release a fork, which is not good IMO :)

felixgabler:
I've tried out the `swipe_to` package [here](https://github.com/flyerhq/flutter_chat_ui/tree/swipe-to-reply-swipe-to) and it definitely works better. However, it just starts an animation on the first movement instead of offering full moving control. This can also be annoying. Please try it out and let me know what you think.

One alternative could be writing a custom widget for this. If you take a look at the source code of swipe to, you'll see that it is not that hard.

Lastly, I changed the swipe direction to reply on swipe right instead of left. This is what most people are most likely used to. Was there a particular reason why it was swipe left? Feel free to change it back then!

demchenkoalex:
@felixgabler swipe right on iOS usually works as a back gesture. Facebook Messenger, Telegram, Instagram all have swipe left actions, not sure why you think swipe right is what people used to :)

Regarding `swipe_to` I believe @Voidozzer checked that, but at the time of implementing this feature, `swipe_to` was unsupported for a year plus and did not work for our case, this was fairly recent update. We will try the `swipe_to` again and see how it looks like. Thanks!

felixgabler:
That's interesting! I checked on WhatsApp, Signal and Teams, which are my most used messengers. I guess it differs for platforms and we can also just choose the other direction. Or perhaps we can make it configurable for the end user. Not really hard with `swipe_to`.

demchenkoalex:
Yeah, should be configurable

Justus-M:
Hi @demchenkoalex! Feels like things have gone a little quiet in this repo. Hope you, your family, and friends are doing well!

Is there a rough expected timeline on the release that will include message replies? Have you decided whether to release it next, or move straight to 2.0?

Happy to help any way I can.

Justus-M:
Ok understood, thanks for the quick reply!

demchenkoalex:
Closing as a duplicate.

PR #256, opened 2022-04-21
Swipe to reply swipeable tile

demchenkoalex:


lib/src/widgets/input.dart
```diff
@@ -114,7 +129,8 @@ class _InputState extends State<Input> {
     final trimmedText = _textController.text.trim();
     if (trimmedText != '') {
       final _partialText = types.PartialText(text: trimmedText);
-      widget.onSendPressed(_partialText);
+      widget.onSendPressed(_partialText,
+          repliedMessage: InheritedRepliedMessage.of(context).repliedMessage!);
```
slozano95, line 133: @demchenkoalex this force cast throws an error when there is no reply message

felixgabler:
This feature would be super awesome to have! Can I help with anything here?

demchenkoalex:
Hi @felixgabler Google I/O is not far away, I thought maybe waiting a bit to see if new Dart APIs or some cool features are introduced, cause right now I don't like this MR, it works, but code is not readable and I would like to refactor it to have small widgets instead of functions, maybe new versions will have some core improvements. WDYT?

felixgabler:
My thought is that yes, the code is not super readable but that is an issue we should perhaps tackle more comprehensively over the whole code base. What I mean to say is that the code quality of this PR does not stand out to me as particularly bad or needing to be fixed.

As for Google I/O, I am not sure that the changes to drop will drastically change what is possible in terms of code refactoring opportunities. The last time I went through the dart spec discussions, it was mostly about enum improvements, better switches (which could be used here but perhaps as a separate PR) and metaprogramming (although that seems a bit more distant). Perhaps it would be best if we opened another issue where we can collect the places that need refactoring to tackle them together. What do you think?

demchenkoalex:
This was the idea yes, to start with this and eventually fix the whole project using small widgets. I wouldn't mind a call if you're available sometime, you definitely have more hands-on experience and together we can make v2 even better in terms of code quality and reusability/customization. I thought about researching a builder pattern for the types, replacing inherited widgets boilerplate with something (there is a lib for that I don't remember the name, it was even mentioned in the Flutter docs if I remember correctly), maybe you will know additional things we can do. I will be closing React Native libraries for now (they distract me) and will be 100% focused on this.

jld3103:
Hi, any updates on this? I would be awesome to have this in the library.

demchenkoalex:
Closing this one for now, the code is on a branch and it too outdated compared to the changes in the repo. Need to find time to finally finish this, it starts to haunt me.

PR #262, opened 2022-05-04
Add bold, italic, strikethrough & code style to input

hareshgediya:
### What does it do?

Format text input message with pattern style.

### Why is it needed?

Text message already using pattern style to format text messages, like. bold, italic, etc.

### Related issues/PRs

Not related to any issues/PRs

### DIFF
--- a/lib/src/widgets/input_text_field_controller.dart
+++ b/lib/src/widgets/input_text_field_controller.dart
@@ -1,0 +1,61 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+
+/// Controller for the [TextField] on [Input] widget
+/// To highlighting the matches for pattern
+class InputTextFieldController extends TextEditingController {
+  /// A map of style to apply to the text pattern
+  final Map<RegExp, TextStyle> _patternStyle = {
+    RegExp('(\\*\\*|\\*)(.*?)(\\*\\*|\\*)'):
+        const TextStyle(fontWeight: FontWeight.bold),
+    RegExp('_(.*?)_'): const TextStyle(fontStyle: FontStyle.italic),
+    RegExp('~(.*?)~'): const TextStyle(decoration: TextDecoration.lineThrough),
+    RegExp('`(.*?)`'):
+        const TextStyle(fontFeatures: [FontFeature.tabularFigures()]),
+  };
+
+  @override
+  TextSpan buildTextSpan({
+    required BuildContext context,
+    TextStyle? style,
+    required bool withComposing,
+  }) {
+    return TextSpan(text: text, style: style).splitMapJoin(
+      RegExp(_patternStyle.keys.map((it) => it.pattern).join('|')),
+      onMatch: (match) {
+        final text = match[0]!;
+        final key = _patternStyle.keys.firstWhere((it) => it.hasMatch(text));
+        return TextSpan(text: text, style: _patternStyle[key]);
+      },
+    );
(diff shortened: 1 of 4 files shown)


lib/src/widgets/input_text_field_controller.dart
```diff
@@ -0,0 +1,61 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+
+/// Controller for the [TextField] on [Input] widget
+/// To highlighting the matches for pattern
+class InputTextFieldController extends TextEditingController {
+  /// A map of style to apply to the text pattern
+  final Map<RegExp, TextStyle> _patternStyle = {
+    RegExp('(\\*\\*|\\*)(.*?)(\\*\\*|\\*)'):
```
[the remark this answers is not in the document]
hareshgediya: I created `PatternStyle` class for that and added static methods so now it will be good.

lib/src/widgets/input_text_field_controller.dart
```diff
@@ -0,0 +1,61 @@
...
+        final style = _listPatternStyle
+            .firstWhere((element) => element.regExp.hasMatch(text))
+            .textStyle;
+        return TextSpan(text: text, style: style);
+      },
+    );
+  }
+}
+
+extension _TextSpanX on TextSpan {
```
felixgabler, line 35: Why build this as an extension to `TextSpan`? This means we have to first instantiate the `TextSpan(text: text)` to the immediately call `toPlainText()` just to get the text. While it looks like an elegant solution, I believe it could be done simpler directly inside the TextEditingController. Also, the `onNonMatch` of this private extension is never used but will construct text span and perform quite a bit of logic. It could be skipped immediately when `onNonMatch` is not supplied

hareshgediya: Maybe it has a simple way, but when I was building for my project at that time I didn't find any simple way except this. Here extension creates for simplicity, to understand logic batter. TextEditingController has a default `buildTextSpan` method for styling and  I'm using that. When the pattern does not match with input text that time it required _default style_ that's specified by `onNonMatch`.

felixgabler: What I meant was that we can just call `text.splitMapJoin` in the `buildTextSpan` and move all the logic in a method there. That would make it better readable.

Regarding the `onNonMatch`, you're right!

hareshgediya: We can't because `text.splitMapJoin` return string. But `buildTextSpan` have return type TextSpan. So first I need to instantiate TextSpan then need to use splitMapJoin.

felixgabler: I don't see how that is a problem. I'm not talking about immedtialey returning the result of the `text.splitMapJoin`. I'm just talking about not overcomplicating things by using the extension method and thereby spreading out logic unnecessarily.

You can just copy the whole content of the extension method into the `buildTextSpan`, modify the `onMatch`, and then you'll return a `TextSpan` as required.

hareshgediya: I got your point. I removed `TextSpan` extension.

felixgabler: Awesome! Code looks good to me

felixgabler (approved):

demchenkoalex (approved):

PR #271, opened 2022-05-27
Add parameter to pass headers when grabbing User's imageUrl

DanMossa:
### What does it do?

You are now able to pass in auth headers when grabbing an image.

### Why is it needed?

The images that are being grabbed require a token to be passed in the header.

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -41,6 +41,7 @@ class Chat extends StatefulWidget {
     this.emojiEnlargementBehavior = EmojiEnlargementBehavior.multi,
     this.emptyState,
     this.fileMessageBuilder,
+    this.headers,
     this.groupMessagesThreshold = 60000,
     this.hideBackgroundOnEmojiMessages = true,
     this.imageMessageBuilder,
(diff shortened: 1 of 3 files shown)


demchenkoalex:
Not sure about this one, first one is the naming? what is `headers` if I don't know the context Chat(headers: ???)? second, we have an image message, can it require headers as well?

DanMossa:
@demchenkoalex All very valid concerns. Do you have any suggestions for the name then?

demchenkoalex:
well it depends, if we do it for all images or for avatars only? if avatars only, something like `avatarNetworkImageHeaders` (I like descriptive names lol) but why we would have it for avatars only.. on the other hand, other network images are in that conditional workarounds in order for chat UI to work on web, not sure if it is easy to pass params there. If it is, we could just have `networkImageHeaders` that go to every network image out there. I guess if header is not user it does not break the fetch.

demchenkoalex:
This became a bit stale, but for anyone having the same issue, I would suggest using `avatarBuilder` and just copy UserAvatar code to your project, and pass headers. This can't be merged as is, too many unknowns why these headers exist for avatars and not other images, and moreover, too specific prop to have on the Chat itself

PR #277, opened 2022-06-13
Add user agent option for preview data fetching

felixgabler:
### What does it do?

Users can now specify a `userAgent` to be sent with the link preview data request.

### Why is it needed?

Many websites send special titles and descriptions when the user agent is e.g. "WhatsApp/2". This is especially relevant if using deep links that heavily depend on the user agent.

### Related issues/PRs


demchenkoalex:
Thanks for the PR!

PR #285, opened 2022-06-29
Add text message options to customize opening links

felixgabler:
### What does it do?

Users can now specify how exactly links should open (e.g. use different launch mode or handle certain links inside their app). I also implemented a first draft options object for text messages which we could work off of.

### Why is it needed?

Some apps have deep links that could be opened immediately without opening another browser.

### How to test it?

### Related issues/PRs

### DIFF
--- a/lib/src/widgets/text_message.dart
+++ b/lib/src/widgets/text_message.dart
@@ -24,6 +24,7 @@ class TextMessage extends StatelessWidget {
     required this.message,
     this.nameBuilder,
     this.onPreviewDataFetched,
+    this.options = const TextMessageOptions(),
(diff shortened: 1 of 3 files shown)


lib/src/widgets/text_message.dart
```diff
@@ -50,6 +51,8 @@ class TextMessage extends StatelessWidget {
   final void Function(types.TextMessage, types.PreviewData)?
       onPreviewDataFetched;
 
+  final TextMessageOptions options;
```
demchenkoalex, line 54: Please add a documentation comment. Otherwise all good 👍

Issue #269, opened 2022-05-24
Other User(s) Typing Indicator


cody1024d:
I think it would be super useful to have specific customization related to showing a typing indicator. I believe the ability is probably there, through the use of a Custom Cell, but it will be a bit hackier/difficult. I believe the framework could supply this ability, and support consistent styling/theming of the typing bubble.

cody1024d:
Want to add a bit more --- the hack I outlined above has the annoying part that it will still try to show an avatar. In my use-case, I want to be able to show a typing indicator, but not show an avatar on that one custom cell.

This is **kind of** possible using the avatarBuilder cutomizable function, however this only exposes the userId, and doesn't allow the fallback of using the default avatar (like I want for my case). Specifically; I want to use the default avatar in all cases BUT the case of the typing message.

demchenkoalex:
This should be a specific widget, without hacks, to be added.

Yahllil:
Any idea when this feature will be available?

malikdoksoz:
> Found some issues in RTL languages, need to fix it, then release.

we've been waiting for this for a long time :)

Lammpo:
Hi! Thank for your amazing chat. Is this feature available yet?

AdrKacz:
Hey 😁 What's the status on this one?

What do you mean by "it is broken on RTL"? What is the behaviour? Moreover, what is the expected behaviour when multiple users are typing?

_Thank you for your amazing job on this package 🎉_

demchenkoalex:
![Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-17 at 22 13 26](https://user-images.githubusercontent.com/14123304/219794453-361b97f1-519a-4dee-b0a4-d405daac92e9.png)

demchenkoalex:
It is working correctly for multiple users typing. No updates here, I am taking a break.

AmirC123:
> Found some issues in RTL languages, need to fix it, then release.

If it's what I think it is, I think I fixed it and opened a pull request 

It sets the writing direction for each message by checking the language (and the input too) so the ! ? etc are all in the correct place

AdrKacz:
Super thank you very much 😁 That would be awesome! Would you need any help?

demchenkoalex:
It is not it, question was about typing indicator and it is broken :)

[...]

PR #293, opened 2022-07-20
add typing indicator

gtalha07:
### What does it do?

This PR implements the typing indicator which is missing from the package. The current change allow Chat widget to add optional typing indicator. It takes `List<types.User>` as a parameter which by default is empty. When empty, it also doesn't show any indicator. When we have more than 2 users in list, it shows small count bubble, otherwise it'll show two users avatars actively typing and indicator. The indicator widget has support of customization in the form of `bubbleBorder`, `bubbleColor`, `animatedCircleColor`,
 `ForwardAnimationDuration` and `ReverseAnimationDuration` to control animation duration of indicator circles. A lot of has been commented as per your pattern in the code. 

### Why is it needed?

This is most useful for people who want to show typing indicator in their chat implementations. I believe it'll help to make package even better. 

### How to test it?

From the example code. We can add users list as a parameter for `showTypingIndicator()`. By default its disabled and takes empty list.

### Related issues/PRs

This PR relates to #269.


demchenkoalex:
Thanks a lot! Will check this out ASAP.

gtalha07:
typingIndicatorDefault
typingIndicatorDark

gtalha07:
Screenshot_1658493904
Screenshot_1658493938

demchenkoalex:
Hey! Thanks again for the big contribution! I hope you don't mind a couple of requests:

1. I asked @dariakhimych our designer to ASAP prepare the design for this feature, I don't think that the typing indicator itself will change much, but after our brief conversation we were on the same page about the multiple users typing thing, that it will be text.
2. To make this library more contribution friendly, together with @felixgabler we decided to minimise amount of props we pass to the `Chat` itself by using 2 things: `Theme` and `Options`. Now `Theme` will be similar to the Flutter's `ThemeData` - you can find tons of other components themes inside `ThemeData`, like `CheckboxThemeData`, `DrawerThemeData`, `SnackBarThemeData` etc. In our case, `ChatTheme` should include `TypingIndicatorTheme`, that will contain all theme customisation for the typing indicator only like colours, borders etc. We don't have an example yet, but I think you can create `TypingIndicatorTheme` class in the same file as the component itself, and make it similar to the `InputOptions` class you can find in the `input.dart`. The other thing, `Options` (might not be applicable to your case) is anything else that is not theme, and not so important to have on the `Chat` itself. See `InputOptions` mentioned above for the example. `showTypingIndicator ` should be on the `Chat` cause it will be widely used, so that is OK.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,284 @@
...
+class TypingIndicator extends StatefulWidget {
+  const TypingIndicator({
+    super.key,
+    required this.author,
+    required this.forwardDuration,
+    required this.reverseDuration,
+    required this.bubbleAlignment,
+  });
+
+  /// See [types.User].
```
felixgabler, line 15: This comment should actually explain what the parameter is for and not it's type

felixgabler (commented):
Looks great!

gtalha07:
Added ```TypingIndicatorTheme``` in ```ChatTheme``` which will allow us to set custom theme for typing indicator. Removed count bubble for multiple users text as per according to your changes. Will look forward for that feature design. Animations for dots is set to ```800ms``` as it doesn't really comply with theme itself and is not important.

demchenkoalex:
[Flyer Chat - Typing.pdf](https://github.com/flyerhq/flutter_chat_ui/files/9193458/Flyer.Chat.-.Typing.pdf)

this is what @dariakhimych prepared. Neutral 1 color is - you can add it to the `chat_theme.dart`. Also feel free to not localise anything, I will add it myself with all the languages. Also pdf includes a link to the example animation our designer is expecting, similar to what you have, maybe different speeds/easings/sequence. 

```dart
const neutral1 = Color(0xff615e6e);
```

gtalha07:
> [Flyer Chat - Typing.pdf](https://github.com/flyerhq/flutter_chat_ui/files/9193458/Flyer.Chat.-.Typing.pdf)
> 
> this is what @dariakhimych prepared. Neutral 1 color is - you can add it to the `chat_theme.dart`
> 
> ```dart
> const neutral1 = Color(0xff615e6e);
> ```
> 
> Also feel free to not localise anything, I will add it myself with all the languages. Also pdf includes a link to the example animation our designer is expecting, similar to what you have, maybe different speeds/easings/sequence.

One thing I want to add, think of it as suggestion. Since the typing indicator takes textStyle for multiple users. Why not we allow it to take as widget. ```Text``` is also a widget but introducing that will allow developers to put any widget inside it. For instance, it'll also allow us to put ```UserAvatar``` instead of text and more accessibility for those who needed to place avatars?

demchenkoalex:
Hey @gtalha07 what will you say if our designer will make an alternative solution with avatars, cause I don't think it will be easy to just give widget and lay out both text or avatars, since text just says "N others" and avatar like you suggested will have +4 or something. You think this is feasible? We will do like a mode "Text" / "Avatar"?

gtalha07:
> Hey @gtalha07 what will you say if our designer will make an alternative solution with avatars, cause I don't think it will be easy to just give widget and lay out both text or avatars, since text just says "N others" and avatar like you suggested will have +4 or something. You think this is feasible? We will do like a mode "Text" / "Avatar"?

Hi @demchenkoalex, I think the possibility is there to make typing indicator toggle between text and avatars. This is just a demo of what we can do for avatars. 
Screenshot_1659008720

YassineChe:
Hello.
Can you please be more specific about this feature implementation, cuz i couldn't find `showTypingIndicator()` attribute.
Or please, show us a snippet,
Thanks!

demchenkoalex:
@YassineChe this is a pull request, where we discuss and finalise code before merging. Then, after a release, you will be able to use it.

demchenkoalex:
Hey @gtalha07, sorry for the delay, here is what our designer made. I can see here 3 modes actually: only text, only avatar or both.

typing-with-avatars

gtalha07:
@demchenkoalex I'm also looking here the support of ```indicatorBuilder``` which can allow to set up custom typing indicator widget. What do you think?

demchenkoalex:
Yes, @gtalha07 agree!

gtalha07:
> Yes, @gtalha07 agree!

@demchenkoalex Please review it. I think its looking good now and cool package needs update. 😄

doc/advanced-usage.md
```diff
@@ -555,3 +555,8 @@ class _MyHomePageState extends State<MyHomePage> {
 ## User avatars & names
 
 To show user avatars & names use `showUserAvatars` and `showUserNames` parameters. Can be used separately. By default, the chat will select one of 10 provided colors as an avatar background and name text color. Color is calculated based on the user's `id` hash code, so it is unique in different rooms. To modify provided colors use `userAvatarNameColors` parameter in [theme](themes). If you want to have one color for everyone, just pass this color as a single item in the `userAvatarNameColors` list.
+
+## Typing Indicator
+To show typing indicator, you can use `showTypingIndicator` optional parameter which takes [User.types] list. By default its empty list which hides the indicator. Shows typing user avatar along with animated speech bubble. You can set your own theme for typing indicator, see [theme](themes).
```
felixgabler, line 560:
```suggestion
To show a typing indicator, you can use the optional parameter `showTypingIndicator` which requires a type of `List<types.User>`. If provided a non-empty list, it will show a typing user avatar along with animated speech bubble. If the list is empty, it will hide the indicator. To set your own theme for the typing indicator, see [theme](themes).
```

What exactly does theme resolve to?

lib/src/chat_theme.dart
```diff
@@ -403,6 +412,19 @@ class DefaultChatTheme extends ChatTheme {
       fontWeight: FontWeight.w800,
       height: 1.333,
     ),
+    super.typingIndicatorTheme = const TypingIndicatorTheme(
+      bubbleColor: neutral7,
```
felixgabler, line 416: sort properties alphabetically

lib/src/widgets/chat.dart
```diff
@@ -3,25 +3,28 @@ import 'dart:math';
 import 'package:flutter/material.dart';
 import 'package:flutter_chat_types/flutter_chat_types.dart' as types;
 import 'package:intl/intl.dart';
-import 'package:photo_view/photo_view.dart' show PhotoViewComputedScale;
```
felixgabler, line 6: You're fixing conflicts commit [3bae598](https://github.com/flyerhq/flutter_chat_ui/pull/293/commits/3bae59834ff96db88158befdc99c029b281a4ff3) actually reverted some progress (i.e. extraction of `ImageGallery` widget and using `VoidCallback`. Please make sure to bring them back.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+  @override
+  void dispose() {
+    _appearanceController.dispose();
+    _animatedCirclesController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    size = MediaQuery.of(context).size;
```
felixgabler, line 85: I'd say its not good practice to retain build time constants like this outside of the build function. Better to pass them through where they are needed.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+            : MainAxisAlignment.end,
+        children: <Widget>[
+          widget.bubbleAlignment == BubbleRtlAlignment.left
+              ? Expanded(
+                  child: Container(
+                    margin: const EdgeInsets.only(right: 12),
+                    child: _typingContainer(widget.indicatorMode),
+                  ),
+                )
+              : Container(),
```
felixgabler, line 104: empty `Container()` can be `const SizedBox()` to stay efficient. Our analysis warnings should report this already. Do you have them enabled?

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+        ],
+      ),
+    );
+  }
+
+  void _showIndicator() {
+    _appearanceController
+      ..duration = const Duration(milliseconds: 500)
```
felixgabler, line 161: It would be easier to set the duration during instantiation.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+      stackingWidth = size.width * 0.11;
+    } else {
+      stackingWidth = size.width * 0.15;
+    }
+    return stackingWidth;
+  }
+
+  /// Display mode of typing.
+  Widget _typingContainer(TypingIndicatorMode mode) {
+    if (mode == TypingIndicatorMode.text) {
```
felixgabler, line 249: A `switch` would be nicer to get IDE warnings when the enum changes.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+      );
+    }
+  }
+
+  /// Handler for multi user typing text.
+  String _multiUserTextBuilder(List<types.User> author) {
+    if (author.isEmpty) {
+      return '';
+    } else if (author.length == 1) {
+      return '${author.first.firstName} is typing';
```
felixgabler, line 287: internationalization required

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+    } else if (author.length == 2) {
+      return '${author.first.firstName} and ${author[1].firstName}';
+    } else {
+      return '${author.first.firstName} and ${author.length - 1} others';
+    }
+  }
+
+  // Typing avatar Widget.
+  Widget _typingAvatar(types.User author) {
```
felixgabler, line 296: You could extract a widget here. The more the better :)

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+  final BubbleRtlAlignment bubbleAlignment;
+
+  /// Used to toggle [TypingIndicator] mode. See [TypingIndicatorMode].
+  final TypingIndicatorMode indicatorMode;
+
+  @override
+  State<TypingIndicator> createState() => _TypingIndicatorState();
+}
+
+class _TypingIndicatorState extends State<TypingIndicator>
```
felixgabler, line 28: This class is very long. Please take a look whether you can extract more widgets. That would also be better for performance.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+              textScaleFactor: 0.7,
+            )
+          : null,
+    );
+  }
+
+  /// Handler for multi user typing avatars.
+  Widget _multiUserAvatarBuilder(List<types.User> author) {
+    if (author.isEmpty) {
+      return Container();
```
felixgabler, line 323:
```suggestion
      return const SizedBox();
```

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+          decoration: BoxDecoration(
+            color: circlesColor,
+            shape: BoxShape.circle,
+          ),
+        ),
+      );
+}
+
+@immutable
+class TypingIndicatorTheme {
```
felixgabler, line 407: @demchenkoalex where do we want to keep the themes? With the widget itself or in the main theme file? For the unseen messages, I placed it in the theme file but I also like this approach.

felixgabler (changes requested):

doc/advanced-usage.md
demchenkoalex: sections on the left side

Screenshot 2022-08-06 at 13 10 00

but docs do not need a review, I will review them myself and re-write if needed.

lib/src/widgets/typing_indicator.dart
demchenkoalex: I thought with the widget, similar to options. Nice to have everything you need in one file.

lib/src/widgets/typing_indicator.dart
gtalha07: I already had analysis checks enabled in my environment yet it didn't announced here. Will change it.

lib/src/widgets/typing_indicator.dart
gtalha07: This was discussed earlier and will be handled by @demchenkoalex himself.

lib/src/widgets/chat.dart
```diff
@@ -81,10 +84,12 @@ class Chat extends StatefulWidget {
     this.scrollToUnseenOptions = const ScrollToUnseenOptions(),
     this.showUserAvatars = false,
     this.showUserNames = false,
+    this.showTypingIndicator = const [],
```
felixgabler, line 87: These options could also be added to an option class similar to `InputOptions`

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,440 @@
...
+    );
+  }
+
+  void _showIndicator() {
+    _appearanceController
+      ..duration = const Duration(milliseconds: 500)
+      ..forward();
+  }
+
+  Animation<Offset> _firstCircleOffset() => TweenSequence<Offset>(
```
gtalha07, line 165: @felixgabler I'm trying to convert this method into reusable one like this
```_circleOffset(Offset? start, Offset? end, double startInterval, double endInterval)```
However we would need to define these as ```const``` before ```build```. Do you have any opinion whether this would be feasible or not?

Comment (gtalha07) [reply lib/src/widgets/typing_indicator.dart:165]: Because we need to pass values to ```_animatedCirclesController``` in ```initState()```, if we go with that.

Comment (felixgabler) [reply lib/src/widgets/typing_indicator.dart:165]: First of all, good idea to make it reusable by parameterisation. It should lead to exactly the same behavior so I don't quite understand what you mean with const? The parameters for it would still remain `const`. Could you maybe send me some code that you think would break here?

Comment (gtalha07) [reply lib/src/widgets/typing_indicator.dart:165]: Sorry not ```const``` but literal values like this:

```
void initState() {
    super.initState();

    //

    _firstCircleOffsetAnimation = _circleOffset(
      Offset.zero,
      const Offset(0.0, -0.90),
      0.0,
      1.0,
    );
    _secondCircleOffsetAnimation = _circleOffset(
      Offset.zero,
      const Offset(0.0, -0.80),
      0.30,
      1.0,
    );
    _thirdCircleOffsetAnimation = _circleOffset(
      Offset.zero,
      const Offset(0.0, -0.90),
      0.45,
      1.0,
    );
    
    //

  }
  ```
  **The method will be:**
  ```
  Animation<Offset> _circleOffset(
    Offset? start,
    Offset? end,
    double startInterval,
    double endInterval,
  ) =>
      TweenSequence<Offset>(
        <TweenSequenceItem<Offset>>[
          TweenSequenceItem<Offset>(
            tween: Tween<Offset>(
              begin: start,
              end: end,
            ),
            weight: 50.0,
          ),
          TweenSequenceItem<Offset>(
            tween: Tween<Offset>(
              begin: end,
              end: start,
            ),
            weight: 50.0,
          ),
        ],
      ).animate(CurvedAnimation(
        parent: _animatedCirclesController,
        curve: Interval(startInterval, endInterval, curve: Curves.linear),
        reverseCurve:
            Interval(startInterval, endInterval, curve: Curves.linear),
      ));
      
   ```

felixgabler: You could also extract the `const Interval(startInterval, endInterval)` since it's used twice. Overall, extracting the method is definitely a code improvement so I'd say go for it!

gtalha07:
@felixgabler Have a look at new code. Many changes have been done. It looks nice thanks to you!

- [ ] Extracted widgets where possible.
- [ ] Refactor methods into reusable one.
- [ ] Added ```typingIndicatorOptions```. This now takes author(s) list by same mechanisam. Removed ```showTypingIndicator```.
- [ ] Proposed changes as per review.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+    return indicatorWidth * 0.06;
+  } else if (author.length == 2) {
+    return indicatorWidth * 0.11;
+  } else {
+    return indicatorWidth * 0.15;
+  }
+}
+
+/// Animation Interval for circles.
+Interval _animationInterval(
```
felixgabler, line 437: I don't think this is really necessary because the two calls from within `_circleOffset` have the same arguments. I'd be in favour of making the third parameter for `_circleOffset` an interval directly. Then, it can also be const.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+class TypingIndicatorOptions {
+  const TypingIndicatorOptions({
+    this.authors = const [],
+    this.animationSpeed = const Duration(milliseconds: 500),
+    this.typingMode = TypingIndicatorMode.text,
+  });
+
+  /// Author(s) for [TypingIndicator].
+  /// By default its empty list which hides the indicator, see [types.User].
+  final List<types.User> authors;
```
felixgabler, line 454: I'm wondering if `authors` really is a good name for this property. I think `typingUsers` would explain the function and type better, what do you think?

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+class _TypingIndicatorState extends State<TypingIndicator>
+    with TickerProviderStateMixin {
+  late double stackingWidth;
+  late AnimationController _appearanceController;
+  late AnimationController _animatedCirclesController;
+  late Animation<double> _indicatorSpaceAnimation;
+  late Animation<Offset> _firstCircleOffsetAnimation;
+  late Animation<Offset> _secondCircleOffsetAnimation;
+  late Animation<Offset> _thirdCircleOffsetAnimation;
+  @override
```
felixgabler, line 32: nit: In general, I'd prefer to leave a space before messages for readability.

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+class AnimatedCircles extends StatelessWidget {
+  const AnimatedCircles({
+    super.key,
+    required this.circlesColor,
+    required this.animationOffset,
+  });
+
+  final Color circlesColor;
+  final Animation<Offset> animationOffset;
+  @override
```
felixgabler, line 392: nit: new line before method

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+          decoration: BoxDecoration(
+            color: circlesColor,
+            shape: BoxShape.circle,
+          ),
+        ),
+      );
+}
+
+/// Handler for multi user typing text.
+String _multiUserTextBuilder(List<types.User> author) {
```
felixgabler, line 413: This method should live in the class where it is used (i.e. `TypingWidget`)

lib/src/widgets/typing_indicator.dart
```diff
@@ -0,0 +1,498 @@
...
+    return '${author.first.firstName} is typing';
+  } else if (author.length == 2) {
+    return '${author.first.firstName} and ${author[1].firstName}';
+  } else {
+    return '${author.first.firstName} and ${author.length - 1} others';
+  }
+}
+
+/// Used to specify width of stacking avatars based on number of authors.
+double _getStackingWidth(List<types.User> author, double indicatorWidth) {
```
felixgabler, line 426: Same here.

felixgabler (commented):
Good work! We're almost there :)

@demchenkoalex only thing I am not sure about (aside from tiny nits) is whether we should move the extracted widgets into their own files. What do you think?

gtalha07:
@felixgabler However with some new changes in [commit](https://github.com/flyerhq/flutter_chat_ui/commit/36e1dd44b735b65c1d8491198a4fcd3d1a31e58c), the typing indicator area seems to be fixed on screen like ```input``` widget. It doesn't scroll along with messages. Do I need scroll controller here?  You can test through ```example```.

felixgabler:
@gtalha07 I checked out your PR at different commits and this behaviour was already there before the commit you mentioned. To get the desired behavior, the typing indicator would need to be a sliver in the `CustomScrollView` in `chat_list.dart`. Is this something we want in the design?

gtalha07:
> @gtalha07 I checked out your PR at different commits and this behaviour was already there before the commit you mentioned. To get the desired behavior, the typing indicator would need to be a sliver in the `CustomScrollView` in `chat_list.dart`. Is this something we want in the design?

It seems like it got changed upon adding code. @demchenkoalex didn't mentioned specifically anything about this but it's expected the widget should be before latest message.

felixgabler:
Then it will have to be moved somewhere else. It can't scroll with the messages if it's not part of the chat list. Right now, it's directly next to the Input and as such also behaves like that, i.e. is fixed

lib/src/widgets/typing_indicator.dart
demchenkoalex: Yep, this is ok! Sorry :)

demchenkoalex:
First of all, thanks @gtalha07 so much for your work and thanks @felixgabler for reviewing this.

Regarding the position of the indicator among other widgets - on videos attached before, indicator correctly appeared and animated all messages to accommodate space for itself. We could observe a similar behaviour in Signal or Messenger. However, if the user scrolls chat even for like 10 pixels (i.e. we are not in the beginning of the scroll list), indicator will not push messages up, so it would behave like it never appeared, but then if you scroll back to the start of the list, it will be visible. So there are 2 cases:

1. List is not scrolled. Indicator appears and animates messages up.
2. List is scrolled for at least 1 pixel, indicator appears in the list, but does not scroll content up, user needs to scroll down to see it.

Does this makes sense? Will it be hard to do?

Another thing I noticed while looking at the videos above, that when indicator is removed there is no animation, it just disappears - will it be possible to keep it alive for a duration of the delete animation and then remove the instance?

gtalha07:
The intentional animation is as follows:
TIAnimation
However I can't seem to call ```_appearanceController.reverse()``` before typing indicator ```dispose()``` method. Because on dispose, it just removes widget without animation even if we call reverse animation there. Any ideas of how to achieve this ? @demchenkoalex @felixgabler.

felixgabler:
> The intentional animation is as follows:
> TIAnimation
> However I can't seem to call ```_appearanceController.reverse()``` before typing indicator ```dispose()``` method. Because on dispose, it just removes widget without animation even if we call reverse animation there. Any ideas of how to achieve this ? @demchenkoalex @felixgabler. 
> 

Either you make sure it does not get disposed until the animation finishes or we keep the typing indicator always alive. The former would be more efficient but probably have worse code because we need to either have callbacks or set two booleans (typing indicator showing and removed) when the typingUser array changes that control it's visibility with removed being delayed until animation finishes.

I'd be okay with always showing it but maybe add a feature flag then for people who are never going to use the typing indicator anyways. Then, they can prevent the minor overhead.

lib/src/widgets/chat_list.dart
```diff
@@ -131,6 +143,25 @@ class _ChatListState extends State<ChatList>
           physics: widget.scrollPhysics,
           reverse: true,
           slivers: [
+            widget.typingIndicatorOptions!.customTypingIndicator != null
```
felixgabler, line 146: The `SliverPadding` and `SliverToBoxAdapter` are the same in both branches so it would be best to pull the ternary operator into there (at the child of `SliverToBoxAdapter`).

Also, you could consolidate the ternary operator using the null-coalescing operator.

```dart
child: widget.typingIndicatorOptions?.customTypingIndicator ?? TypingIndicator(..),
```

gtalha07:
@demchenkoalex Are you referring to this behavior on scrolling content and typing indicator status?  
https://user-images.githubusercontent.com/68579938/185659347-e406cc4f-b608-49b0-960c-708db9876f0b.mp4

demchenkoalex:
will resolve this ASAP. I want to merge this and if I will see that something is not right I will try to fix it myself. Sorry for the waiting and thanks for the huge amount of work included here!

elhe26:
> will resolve this ASAP. I want to merge this and if I will see that something is not right I will try to fix it myself. Sorry for the waiting and thanks for the huge amount of work included here!

Thanks, @demchenkoalex .  Is there any roadmap to follow?

demchenkoalex:
Hey @elhe26, the roadmap is straightforward - finish swipe to reply and then all highly requested features - reactions, audio, video messages, time inside bubbles. Everything else is a bonus :)

ramy9109:
any update coming soon for typing Indicator?

demchenkoalex:
yes, I cleaned up tons of issues and MRs last week, so I can do this and replies now.

demchenkoalex:
Thanks for the PR!

demchenkoalex:
Just wanted to add @gtalha07 after all discussions this works perfectly how I imagined it to work, awesome job and sorry for the delay.

rlee1990:
Any roadmap on when this will be live?

baptistesx:
@demchenkoalex any news on the availability of this nice feature? Thank you guys for your awesome work!

demchenkoalex:
why roadmap? it is already merged if you use the lib from GitHub directly, and to release on pub.dev I need to fix RTL languages, didn't find time yet.

baptistesx:
I see, my bad i wasn't understanding as i saw the merge but the feature wasn't available on pub, thanks!

baptistesx:
I got a quick question: how can we trigger the user is typing (to emit the event on the chat socket)? Any way to access the textfield onChange on something similar?

prkhrv:
Is this feature available on pub?

iqfareez:
@prkhrv I think not yet. But hopefully they can publish to pub asap

thomers:
Hello, I'd like to show a typing indicator like this:

Screenshot 2024-06-29 at 11 10 48

Is there documentation on this? What do I have to do? Thanks! Thomas

demchenkoalex:
start with `typingIndicatorOptions` param of the `Chat` and pass `TypingIndicatorOptions` class there. Pass users that are typing to the `typingUsers` array and the loading indicator will appear. Play with other params in `TypingIndicatorOptions` to achieve required style.

thomers:
I'm trying to show the avatar left from the typing indicator - what I get is this:

IMG_C20C9028BB0F-1

The white background is somehow too small for the animation. 

Here's my code.  

` theme: const DefaultChatTheme(
                  seenIcon: Text('read', style: TextStyle(fontSize: 12.0)),
                  inputBackgroundColor: Colors.white12,
                  backgroundColor: Colors.transparent,
                  inputTextColor: Colors.white,
                  primaryColor: chatPrimaryColor,
                  secondaryColor: chatSecondaryColor,
                ),
                typingIndicatorOptions: TypingIndicatorOptions(
                  typingMode: TypingIndicatorMode.avatar,
                )`


How can I fix the background and show the avatar left from the animation? 

Thomas

PS: Thanks for this awesome package! Really helps a lot.

demchenkoalex:
not sure showing avatar left from animation is possible; If you have multiple people typing your animation might go out of the screen. White background seems like a bug, it should not have any background right?

PR #296, opened 2022-07-31
Avoid unnecessary rebuilds of message widgets, allow selective State KeepAlive

### LINKED ISSUE #287: How to prevent chat rebuild when adding new messages?
I have a customMessageBuilder, that's used to build video messages. The problem is that when I add a new message, the entire page/chat is being rebuilt, therefore my custom video player messages are being rebuilt as well

image

https://user-images.githubusercontent.com/16257821/176899647-c07df75a-8686-4019-81aa-2b35d5cd85ae.mp4

otto-dev:
Fixes #287

Uses `SliverList` with `findChildIndexCallback` to avoid rebuilds of messages and retain State.

Animates new messages using a `AnimatedMessage` wrapper that performs the same animation that was previously provided through `SliverAnimatedList`.

Previously "seen" messages are not animated. We track seen messages in a `Set` of `messageIds`. The initially provided messages are all considered "seen". Any messages added later are considered un-seen.

Custom messages can extend flutter's `AutomaticKeepAliveClientMixin` and `@override bool get wantKeepAlive => true;` to selectively retain their state _**even if they scroll out of view**_. Without this setting, message widgets will still _not_ be rebuilt as long as they are in view, but will be destroyed otherwise.

Why `SliverList` instead of `SliverAnimatedList`? Simply because `SliverAnimatedList's` `findChildIndexCallback` appears to be broken. 

Only functional difference that I'm aware of is that messages that disappear are no longer animated away. But this is a very rare and unusual scenario (usually, deleted messages are replaced with a placeholder rather than just "disappear")

You can see an example below with:
- Retained state while a message is in view
- Retained state for a messages out of view if KeepAlive is requested
- Animation of new messages as before (hard to see in the low fps gif, slow motion below, or compile the linked example code)

Record_example_20220731161042

Example code: https://github.com/otto-dev/flutter_chat_ui/blob/example/example/lib/main.dart

Slow motion:
Record_example_20220731163541

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -439,7 +439,6 @@ class _ChatState extends State<Chat> {
               : min(constraints.maxWidth * 0.78, 440).floor();
 
       return Message(
-        key: ValueKey(message.id),
         avatarBuilder: widget.avatarBuilder,
         bubbleBuilder: widget.bubbleBuilder,
         bubbleRtlAlignment: widget.bubbleRtlAlignment,
--- a/lib/src/widgets/chat_list.dart
+++ b/lib/src/widgets/chat_list.dart
@@ -59,6 +59,7 @@ class ChatList extends StatefulWidget {
 /// [ChatList] widget state.
 class _ChatListState extends State<ChatList>
     with SingleTickerProviderStateMixin {
+  final Set<String> seenIds = {};
   late final Animation<double> _animation = CurvedAnimation(
     curve: Curves.easeOutQuad,
     parent: _controller,
@@ -67,9 +68,6 @@ class _ChatListState extends State<ChatList>
   late final AnimationController _controller = AnimationController(vsync: this);
 
   bool _isNextPageLoading = false;
-  final GlobalKey<SliverAnimatedListState> _listKey =
-      GlobalKey<SliverAnimatedListState>();
-  late List<Object> _oldData = List.from(widget.items);
   late ScrollController _scrollController;
 
   @override
@@ -77,6 +75,11 @@ class _ChatListState extends State<ChatList>
     super.initState();
 
     _scrollController = widget.scrollController ?? ScrollController();
+    for (final item in widget.items) {
+      _mapMessage(item, (message) {
+        seenIds.add(message.id);
+      });
+    }
     didUpdateWidget(widget);
   }
 
@@ -84,7 +87,7 @@ class _ChatListState extends State<ChatList>
   void didUpdateWidget(covariant ChatList oldWidget) {
     super.didUpdateWidget(oldWidget);
 
-    _calculateDiffs(oldWidget.items);
+    _scrollToBottomIfNeeded(oldWidget.items);
   }
 
   @override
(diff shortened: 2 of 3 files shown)


lib/src/widgets/chat.dart
```diff
@@ -439,7 +439,6 @@ class _ChatState extends State<Chat> {
               : min(constraints.maxWidth * 0.78, 440).floor();
 
       return Message(
-        key: ValueKey(message.id),
```
otto-dev, line 442: The `SliverList` builder is in control of the `Keys` now, so that it can utilize `findChildIndexCallback`

lib/src/widgets/chat_list.dart
```diff
@@ -134,11 +137,34 @@ class _ChatListState extends State<ChatList>
...
+                    final item = widget.items[index];
+                    final animate = _mapMessage(
+                          item,
+                          (message) => seenIds.add(message.id), // returns true if item not yet in the set
+                        ) ??
+                        false;
+                    return AnimatedMessage(
+                      key: _valueKeyForItem(item),
+                      child: widget.itemBuilder(item, index),
+                      animate: animate,
```
otto-dev, line 152: We use an `AnimatedMessage` wrapper even if there is no animation to be performed (`animated = false`), because switching to a different wrapper would trigger a rebuild of the wrapped child.

lib/src/widgets/chat_list.dart
```diff
@@ -134,11 +137,34 @@ class _ChatListState extends State<ChatList>
...
+                    if (key is ValueKey<Object>) {
+                      final newIndex = widget.items.indexWhere(
+                        (v) => _valueKeyForItem(v) == key,
+                      );
+                      if (newIndex != -1) {
+                        return newIndex;
+                      }
+                    }
+                    return null;
+                  },
```
otto-dev, line 165: This is the magic that allows `SliverList` to retain the state of its children. Without this function, messages will be rebuilt. 

Also available for `SliverAnimatedList`, but appears to be broken upstream.

pubspec.yaml
```diff
@@ -11,7 +11,6 @@ environment:
   flutter: ">=2.0.0"
 
 dependencies:
-  diffutil_dart: ^3.0.0
```
otto-dev, line 14: We don't need this anymore, we only check for new messages, using a `Set` of `messageIds` - also much more performant.

demchenkoalex:
Hello! Thanks a lot for this! I would probably never use `SliverAnimatedList` in the first place, cause I used simple list and added an animation to the message itself, and I was happy until I tried to _delete a message_. This is where I cracked my head for weeks and eventually came up with this solution (to calculate diffs and animate automatically). See for yourself:

|Before|After|
|-|-|
|before|after|

I did not find a way to retain animation in message, after it was deleted (to animate it). If there is a way to do that, I would love to learn that, otherwise, I would trade this much needed optimisation for user experience any day.

otto-dev:
Hi, it's possible to implement a deletion animation based on this PR. You would do it like this:

Keep diffing the message IDs like you used to do with diffutil. When a message is deleted, don't remove the message from the ChatList state, but add the ID to a set of `deletedIds` (just like we have for `seenIds`). Pass a new boolean `deleted` to `AnimatedMessage`. If the `AnimatedMessage` is currently instantiated (i.e. `initState` has been called previously), animate the message out. If an `AnimatedMessage` is instantiated with `deleted == true`, render nothing.

demchenkoalex:
If you can that would be awesome! I will merge it then :)

demchenkoalex:
Can we also remove those "deleted" messages from the memory after some delay? To not have normal and deleted messages for the lifetime of the app

otto-dev:
I tracked down the issue with `SliverAnimatedList` upstream. https://github.com/flutter/flutter/pull/108710

I don't know how much red tape awaits me to merge this upstream, so I might make a separate PR here.

demchenkoalex:
Hey @otto-dev I just realised you proposed 2 different solutions (I thought second MR was adding delete animation). What do you think will be a preferred solution? This one with eventually delete animation or the fixed animated list?

otto-dev:
This PR is now superseded by #298.

The only reason why I replaced `SliverAnimatedList` with `SliverList` is because `indexToItemIndex` is broken in Flutter for `SliverAnimatedList`. But I [fixed that upstream](https://github.com/flutter/flutter/pull/108710), so all we have to do is add `indexToItemIndex` like I did in my other PR, and we can keep the current animations. That is, once my upstream PR is merged and a new flutter version released.

The other PR contains a fixed version of Flutter's `SliverAnimatedList`. We can just remove that and wait for the Flutter release instead

demchenkoalex:
Thanks so much for this @otto-dev!

PR #297, opened 2022-07-31
widgets/Chat add onCloseGalleryPressed/0; ImageGalleryOptions add sho…

leeyisoft:
…wGalleryCloseButton; ImageGallery add "GestureDetector/onTap: () => onClosePressed()"

### What does it do?

Describe the technical changes you did.

### Why is it needed?

Describe the issue you are solving.

### How to test it?

### Related issues/PRs

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -84,7 +84,10 @@ class Chat extends StatefulWidget {
     this.usePreviewData = true,
     required this.user,
     this.userAgent,
+    this.onCloseGalleryPressed,
   });
+
+  final void Function()? onCloseGalleryPressed;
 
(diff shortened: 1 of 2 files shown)


demchenkoalex:
Hi, thanks for the PR! I will have to close it, cause I don't see any value here.

1. `onMessageDoubleTap` change - this should not be here at all. If you want to open gallery on message double tap (strange IMO) just use that callback in your code.
2. `onCloseGalleryPressed` does not handle press "on gallery close", it just closes the gallery on any touch, which does not make sense, since you will not be able to swipe in the gallery, resize images etc.
3. You can always disable image gallery completely, `disableImageGallery`, and re-do as you like.

PR #304, opened 2022-08-06
Fix scroll to index

felixgabler:
### What does it do?

Keep track of scroll indices by message id to allow jumping to them.

### Why is it needed?

I was not aware that the indices for `scroll_to_index` need to be ordered. Had to replace the hash code with a mapping.

### How to test it?

Now you can use the latest message id as `lastSeenMessageID` and it should jump to it.

### Related issues/PRs


demchenkoalex:
Thanks for the MR!

PR #303, opened 2022-08-06
Code improvements

felixgabler:
### What does it do?

- Extracted files into folders for better overview (open to change folders)
- Move theme for unseen messages into widget file
- Little bit of code changes

### Why is it needed?

Better structure makes it easier for new people to navigate the code base.

PR #305, opened 2022-08-07
Display system messages

felixgabler:
### What does it do?

Show system messages for things like 'YXZ joined the chat'.

### Why is it needed?

Using custom messages for such use cases is not great because they are not centered. This makes system messages a first class concept.

### Related issues/PRs

[Types PR](https://github.com/flyerhq/flutter_chat_types/pull/26#issuecomment-1203012479)

Built on top of https://github.com/flyerhq/flutter_chat_ui/pull/303. Wait for it before you review and merge.


demchenkoalex:
Thanks for the rebase! I am not sure about this implementation, I think system message should be part of the animated list like any other message? Otherwise it will just drop in randomly.

felixgabler:
> Thanks for the rebase! I am not sure about this implementation, I think system message should be part of the animated list like any other message? Otherwise it will just drop in randomly.

They are part of the animated list though. The `Message` widget does not perform animation critical functions but just provides avatar, name, border radius etc. For system messages, we really want to prevent all of that and place it centered.

The only things that we could think about adding is gesture detection (although I don't know how useful that would be) and maybe visibility detection (also not sure how useful). In any case, fitting the system messages into the constraints of the `Message` widget would be a mistake imho.

demchenkoalex:
OK, my mistake :D

felixgabler:
Hey @demchenkoalex, is there anything missing before we can merge?

demchenkoalex:
I don't think so, will take a look

demchenkoalex:
will resolve this ASAP

demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #331, opened 2022-10-08
adds bottomWidget to ChatList

MaddinMade:
### What does it do?

I just added the ability to put a custom widget at the start of the `ChatList`.
This pull request is not breaking in any way, it just adds another parameter that has no effect on the behavior of the `Chat`.

### Why is it needed?

You can only add custom widgets to the list so far by implementing a custom message for it and then using the `customMessageBuilder`, which is relatively complicated.


demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

PR #332, opened 2022-10-08
fixes unnecessary scrolls to the bottom

MaddinMade:
### What does it do?

Chat now only scrolls to the new message if it is really new (different ID) and not if it has changed in any way.

### Why is it needed?

Until now, you are always sent to the beginning of the list if the first message in the chat has changed just a little. E.g. if the status of the message changes or if you slightly adjust other things afterwards (e.g. because you synchronize the message with the backend).


demchenkoalex:
Thanks for the PR!

Issue #334, opened 2022-10-10
UTC support


marinkobabic:
Server returns the MillisecondsSinceEpoch but it uses UTC on the server. So we need a new option to enable utc on the client to display the proper time. Please see therefore DateTime.fromMillisecondsSinceEpoch https://api.dart.dev/be/169657/dart-core/DateTime/DateTime.fromMillisecondsSinceEpoch.html

demchenkoalex:
MR was merged, I will try to make use of DateTime going forward (and replacing old dates)

PR #335, opened 2022-10-11
pass new parameter for utc support

marinkobabic:
https://github.com/flyerhq/flutter_chat_ui/issues/334

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -105,6 +105,7 @@ List<Object> calculateChatMessages(
   DateFormat? dateFormat,
   required int dateHeaderThreshold,
   String? dateLocale,
+  bool dateIsUtc = false,
   required int groupMessagesThreshold,
   String? lastReadMessageId,
   required bool showUserNames,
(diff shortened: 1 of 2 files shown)


jld3103:
Maybe it's better to just pass a `DateTime` to the library in the first place? I found this very confusing at first since how it's right now is not really a common pattern.

marinkobabic:
I agree. This is what I'm normally facing in other projects instead of the epoch time in milliseconds. It may be hard to introduce this change because of backward compatibility.

jld3103:
Hm yes, but also it's not that huge I imagine and a very good update on the quality of the library.

marinkobabic:
Any comments on this by the owner of the library?

marinkobabic:
@demchenkoalex any way to merge this change as well? Thanks in advance

demchenkoalex:
Hi, I just started to look at the repo :) I will get here. From a quick glance I would prefer `DateTime` solution, so maybe I will do one release with smaller things I will find and the other one, breaking, using `DateTime`.

demchenkoalex:
OK, merging this, but I will be in favour of releasing `DateTime` breaking change later. Thanks for the PR!

marinkobabic:
thats clear, thx :-)

PR #337, opened 2022-10-13
Add ability to specify additional text matchers for text messages

jld3103:
### What does it do?

It adds a parameter to `TextMessage` to allow specifying additional text matchers.

### Why is it needed?

I need to display custom mentions in my text messages using widgets.

### DIFF
--- a/lib/src/widgets/message/text_message.dart
+++ b/lib/src/widgets/message/text_message.dart
@@ -26,6 +26,7 @@ class TextMessage extends StatelessWidget {
     required this.showName,
     required this.usePreviewData,
     this.userAgent,
+    this.textMatchers = const [],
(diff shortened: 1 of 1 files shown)


demchenkoalex (approved):

demchenkoalex:
Thanks for the PR!

demchenkoalex:
Moved `textMatchers` to the `TextMessageOptions` class (otherwise you can't access this) and renamed to `matchers`.

jld3103:
Thanks for merging, I had the matchers on the message level because I need to match based on some metadata of the message
. Maybe we could bring it back and still make it accessible from the options too? Otherwise this feature is pretty useless to me :(

PR #339, opened 2022-10-19
Feature/audio message builder

marinkobabic:
Instead of discussing now for months we could just initially provide an option to override the audio message rendering. It has no default implementation yet and fits to the chat types pull request https://github.com/flyerhq/flutter_chat_types/pull/30 

At some point later a default implementation can be provided using a plugin concept.


demchenkoalex (approved):

demchenkoalex:
Thanks for the MPR

PR #341, opened 2022-10-29
Allow specifying text message matchers per message

jld3103:
### What does it do?

Add fields to allow specifying additional matchers per text message.

### Why is it needed?

See https://github.com/flyerhq/flutter_chat_ui/pull/337#issuecomment-1295803184

### Related issues/PRs

https://github.com/flyerhq/flutter_chat_ui/pull/337

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -524,7 +524,7 @@ class ChatState extends State<Chat> {
       final message = map['message']! as types.Message;
 
       final Widget messageWidget;
-      
+
       if (message is types.SystemMessage) {
         messageWidget = widget.systemMessageBuilder?.call(message) ??
             SystemMessage(message: message.text);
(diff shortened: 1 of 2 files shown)


demchenkoalex:
I don't understand still, how do you pass these matchers? you render `Chat` component and there is nothing passed to chat except `TextMessageOptions`.  Do you somehow render custom text message and use text message from the package?

jld3103:
Ah sorry, I'm using the `textMessageBuilder` to do this.

demchenkoalex:
so that means you do something like `textMessageBuilder` and then you put there `TextMessage` from the lib right? And you're saying that matchers go based on metadata. So in the `textMessageBuilder` you have access to the metadata and the `TextMessage` widget, so why options does not work? you just create options based on the `metadata`, pass matchers there and it works, no?

jld3103:
Oh I think I messed up. I just thought the options were coming from a layer above where I dom't have access to the message itself. My bad.

PR #342, opened 2022-10-29
Fix scroll to unread when no unread messages exist

jld3103:
### What does it do?

Check if the unread header exists before trying to scroll to it. Previously settting `scrollOnOpen: true` when there were no unread messages led to a null check error.

### Why is it needed?

Stop throwing errors.


demchenkoalex:
Thanks for the PR!

PR #378, opened 2022-12-27
Make messages start from top

Anas-Qasem:
**state: ready to be merged**

Hello 👋, I'm using your amazing Package To create a chat in my app.

This PR is not to fix the issue but adding a new feature
 I was facing a problem, if the chat is new The messages start from the bottom usually in famous chatting apps like (WhatsApp, Messenger, etc..,). They start from the top so I added this feature by just adding a simple flag in the chat Widget, this flag changes the value of shrinkWrap in CustomScrollView and makes the column main axis alignment ->  MainAxisAlignment.space-between if it is true

before :
before

after : 
after


mingxin-yang:
I need this feature too!!!!

demchenkoalex:
Sorry for basically ignoring and thanks for the contribution. I am re-writing the whole thing as `v2` with more knowledge and better in every way, and this will be a default behaviour there. Sorry once again.

imagination24:
you are my hero！This solved my problem

PR #399, opened 2023-02-14
update example/pubspec.yaml

ji-seung-ryu:
### What does it do?

This PR is just for updating example/pubspec.yaml.
My flutter version is 3.7.0.
Before fixing, i had an error about this.


omerhertz:
I had the same problem. Please merge this

PR #477, opened 2023-07-19
Fixed repaint if type indicator don't use

mozomig:
#449 issue

### What does it do?

This PR remove type indicator widget from tree when it is not show

### Why is it needed?

We prevent render type_indicator widget when on don't use

### How to test it?

You can open example screen, and open dev tools performance

### Related issues/PRs

[Let us know if this is related to any issue/pull request.](https://github.com/flyerhq/flutter_chat_ui/issues/449)


demchenkoalex:
Thanks for the PR!

PR #512, opened 2023-10-18
added theming options for text input

claudius-kienle:
Added two new theming variables to define an elevation and surface tint color for the input widget.

This allows to highlight the input text slightly, in line with the material ui guidlines.


demchenkoalex:
Thanks for PR!

PR #516, opened 2023-10-31
Add mounted check before animation

elihaialgoaitech:
Prevent an exception 
`AnimationController methods should not be used after calling dispose.` 
While user pop out the chat while animation is in progress


demchenkoalex:
Thanks for the PR!

Issue #514, opened 2023-10-28
Remain characters being converted when press enter key with Japanese input.


okano4413:
Hi.

When I build  flutter_chat_ui app as web application, some Japanese characters being converted remains in TextFiled.  

Here is gif.    

![remain_converted_character_2](https://github.com/flyerhq/flutter_chat_ui/assets/10623619/4bc3972c-3abf-4b0d-92fb-fd2483d16c36)

Here is flutter doctor  summary.  
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.10.3, on macOS 13.5 22G5038d darwin-x64, locale ja-JP)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✗] Xcode - develop for iOS and macOS
    ✗ Xcode installation is incomplete; a full installation is necessary for iOS and macOS development.
      Download at: https://developer.apple.com/xcode/download/
      Or install Xcode via the App Store.
      Once installed, run:
        sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer
        sudo xcodebuild -runFirstLaunch
    ✗ CocoaPods not installed.
        CocoaPods is used to retrieve the iOS and macOS platform side's plugin code that responds to your plugin usage on the Dart side.
        Without CocoaPods, plugins will not work on iOS or macOS.
        For more info, see https://flutter.dev/platform-plugins
      To install see https://guides.cocoapods.org/using/getting-started.html#installation for instructions.
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.2)
[✓] VS Code (version 1.79.2)
[✓] Connected device (2 available)
[✓] Network resources

! Doctor found issues in 1 category.
```  

Result of my investigation, this behavior only happened when we build flutter_chat_ui as web application, and the implementation  of input.dart::_InputState::_inputFocusNode cause this issue.  
I thought, adding check logic of TextEditingController.value.isComposingRangeValid fix the behavior.  
this is my suggestion.  
```
class _InputState extends State<Input> {
  late final _inputFocusNode = FocusNode(
    onKeyEvent: (node, event) {
      if (event.physicalKey == PhysicalKeyboardKey.enter &&
          !HardwareKeyboard.instance.physicalKeysPressed.any(
            (el) => <PhysicalKeyboardKey>{
              PhysicalKeyboardKey.shiftLeft,
              PhysicalKeyboardKey.shiftRight,
[...]

Issue #476, opened 2023-07-19
Japanese input is wrong in Flutter web.


fugumap:
### Japanese input is wrong.
I use it in Flutter web.
(It's working fine on android.)

This is typing "こんにちわ(ko,nn,ni,ti,wa)".
![こんにちは](https://github.com/flyerhq/flutter_chat_ui/assets/121592556/e1154b4f-5aa5-41ee-b020-bff5f3bcd10d)

demchenkoalex:
So if it is working fine on Android, but does not work in web, means Flutter issue?

fugumap:
I see that it works fine on the web with other libraries.
So I don't think it is a Flutter issue.

demchenkoalex:
You can always check the code as well, here is the input https://github.com/flyerhq/flutter_chat_ui/blob/main/lib/src/widgets/input/input.dart#L161 I don't see anything that will jeopardise any particular language. Maybe text controller that tries to format the input? you can supply your own controller via `inputOptions` class https://github.com/flyerhq/flutter_chat_ui/blob/main/lib/src/widgets/input/input.dart#L248 see if this helps?

PR #517, opened 2023-10-31
Japanese input bug

okano4413:
### What does it do?

Fix bug related on convertible character input issue like Japanese.

### Why is it needed?
I think , some users who using CJK input are in trouble by this behavior.

#### How to reproduce?
1. Build  example project of flutter_chat_ui with web browser.
2. Enter Japanese characters by Roman alphabets. for example, こんにちは(ko nn ni ti ha)
3. If you entered above characters, you can see  "kおんにちは" (#476 )
4. Press enter key once, then "kおんにちは" is sent as message, and "おんにちは" remains in text field.( #514)

My commit fix these issues.

### Related issues/PRs
#514 
#476

### DIFF
--- a/lib/src/widgets/input/input.dart
+++ b/lib/src/widgets/input/input.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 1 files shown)


demchenkoalex:
Thanks for the PR!

Issue #450, opened 2023-05-22
Can I increase the chat bubble width so it's more than 440?


rutaba1:
Hi, I am using this plugin on flutter web and I want to extend the chat bubble to full width of the screen if text message is long. However I noticed that chat container width never gets passed 440. Is there a way I can achieve full width for long messages?

demchenkoalex:
I think it is hardcoded somewhere, you can open a PR with a parameter to change that

KingUgochukwu:
@rutaba1 were you able to find where this is hard coded?

demchenkoalex:
possible in `v2`.

PR #528, opened 2023-11-24
#450 update: add message min width

bobz392:
### What does it do?
add a messageMinWidth option in the chat theme configure

### Why is it needed?
if we run the App in macOS, the max message width is still 440.  looks very unusual.

Describe the issue you are solving.

### How to test it?
set messageMinWidth of the chat theme to [double. infinity] and run it on macOS.

### Related issues/PRs
https://github.com/flyerhq/flutter_chat_ui/issues/450

### DIFF
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -80,6 +80,7 @@ abstract class ChatTheme {
     required this.messageBorderRadius,
     required this.messageInsetsHorizontal,
     required this.messageInsetsVertical,
+    required this.messageMinWidth,
     required this.primaryColor,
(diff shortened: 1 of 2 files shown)


bobz392:
I think there are additional jobs that we can do.  like macOS apps, we can change the window size of the app.

So, do we need to redraw the message cell or not?

demchenkoalex:
Thanks for the PR! I am merging this, but v1 is not really extendable and became just a mess of the params for everything. I am heavily working now on v2 and I actually testing on macOS and web, so it will feel more natural on those platforms and mobile will follow. Plus the code will be easier to customise/extend.

PR #532, opened 2023-12-06
Adds l10n for Persian language (fa locale).

xclud:
### What does it do?

Added i10n class for Persian language. The standard locale for Persian is `fa` and `fa-IR` is the most used one.


demchenkoalex:
thanks for the PR!

Issue #496, opened 2023-09-10
How to change overall chat padding?


SanginiG:
<img width="129" alt="Screenshot 2023-09-10 at 6 53 43 PM" src="https://github.com/flyerhq/flutter_chat_ui/assets/121866209/91eecda5-45e9-4d3e-8908-03bf4ae7ae40">
I am unable to change the default padding on the left side of my avatars

demchenkoalex:
I don't think this was added, you can open a PR though

demchenkoalex:
yeah as I replied above, I think it is hardcoded - you can open a PR though 👌

PR #585, opened 2024-04-06
Add support for overriding bubbleMargin using ChatTheme

ishchhabra:
### What does it do?

This change adds support for overriding the bubbleMargin property using ChatTheme. 

### Why is it needed?

This allows users of the package to customize the margin of the chat bubbles, enabling scenarios where the user wants the message to take the entire width of the screen.

### DIFF
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -320,6 +320,7 @@ class DefaultChatTheme extends ChatTheme {
     super.attachmentButtonIcon,
     super.attachmentButtonMargin,
     super.backgroundColor = neutral7,
+    super.bubbleMargin,
     super.dateDividerMargin = const EdgeInsets.only(
       bottom: 32,
       top: 16,
@@ -493,6 +494,7 @@ class DarkChatTheme extends ChatTheme {
     super.attachmentButtonIcon,
     super.attachmentButtonMargin,
     super.backgroundColor = dark,
+    super.bubbleMargin,
     super.dateDividerMargin = const EdgeInsets.only(
       bottom: 32,
       top: 16,
--- a/lib/src/chat_theme.dart
+++ b/lib/src/chat_theme.dart
@@ -59,6 +59,7 @@ abstract class ChatTheme {
     required this.attachmentButtonIcon,
     required this.attachmentButtonMargin,
     required this.backgroundColor,
+    this.bubbleMargin,
     required this.dateDividerMargin,
     required this.dateDividerTextStyle,
     required this.deliveredIcon,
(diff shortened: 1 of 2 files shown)


qeepcologne:
very similar to what i did. Sorry i did not see your work, because it does not mention the corresponding task #496

ishchhabra:
> very similar to what i did. Sorry i did not see your work, because it does not mention the corresponding task #496

My bad, I should've mentioned the open issue in my PR to avoid duplicated work. I think, though, the ability to completely override the entire margin offers higher flexibility and could be a better approach. What do you think @qeepcologne?

ishchhabra:
> I think you have no access to bubbleRtlAlignment in ChatTheme. If you want the same behaviour just with less padding how to do that?

Since `bubbleRightAlignment` is a property that is provided by the user here https://github.com/flyerhq/flutter_chat_ui/blob/d559096583b07a20c08785d5acee6fb016460b48/lib/src/widgets/chat.dart#L43
I think the user can simply access the value that they are passing to `Chat`. If a user is overriding the base functionality, I think it is reasonable to ask the user to completely handle that themselves. Alternatively, to achieve what I think you are trying for (to only modify the left padding), `bubbleMargin` could be modified to rather be `EdgeInsets Function({required EdgeInsets defaultBubbleMargin})` or the package can simply expose `getDefaultBubbleMargin` as a util function. I prefer the latter as that still keeps the functionality as simple as possible.

qeepcologne:
EdgeInsets is well known, clear and flexible.

But if you want the full width you have to do some caculations:
e.g.:
messageWidthRatio: 1 - 0.06 - 2 * 0.02
when the horizontal padding on left+right is MediaQuery.of(context).size.width * 0.02
and avatars are not used.
But thats another topic.

ishchhabra:
> EdgeInsets is well known, clear and flexible.
> 
> But if you want to use the full width you have to do some caculations: e.g.: messageWidthRatio: 1 - 0.06 - 2 * 0.02 when the horizontal padding on left+right is MediaQuery.of(context).size.width * 0.02 and avatars are not used. But thats another topic.

Yep, for now, I think the current implementation should be sufficient. Let me know if you want me to update this PR to also export a `getDefaultBubbleMargin` function or change `bubbleMargin` to a function which takes `defaultBubbleMargin` as parameter to consolidate both our PRs into one.

qeepcologne:
Sorry for the late reply, i tried to switch to your branch today:
- can we use EdgeInsetsGeometry to allow EdgeInsetsDirectional also?

ishchhabra:
@demchenkoalex Could you please review this pull request? Are there any additional changes required before it can be merged?

theprojectfactory:
Hi @demchenkoalex,

Sorry for the ping.

When can we get this released? Our app is needing this exact feature

ishchhabra:
@theprojectfactory In the mean time until there is a new release, you can also directly target the exact code you want using 

```
flutter_chat_ui:
    git: 
      url: ...
      ref: ...
```

demchenkoalex:
just came back from PTO, will release this week

qasim90:
@demchenkoalex Is this feature available in v2?

PR #586, opened 2024-04-08
#496 move hardcoded left padding to theme

qeepcologne:
see #496, current implementation has hardcoded left padding (20). Moved to theme to allow uses to change or remove it (with the current value as default).


qeepcologne:
close in favor of #585

PR #625, opened 2024-08-12
Update to set a custom value for the gap between message groups

seunghyeoks:
### What does it do?

Allows developers to arbitrarily set the spacing between message groups

### Why is it needed?

Previously, the value was fixed at 12, making it impossible to set the value arbitrarily

### How to test it?

You can test this by simply setting the value in Chat as shown below
```dart
body: Chat(
    messagesSpacerHeight: 3,
    ~~~~~
)
```

### Related issues/PRs

None

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -111,6 +111,7 @@ List<Object> calculateChatMessages(
   bool dateIsUtc = false,
   String? dateLocale,
   required int groupMessagesThreshold,
+  double messagesSpacerHeight = 12,
   String? lastReadMessageId,
   required bool showUserNames,
   DateFormat? timeFormat,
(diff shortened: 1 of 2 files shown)


demchenkoalex:
hey, I just released last version of `v1` if you don't mind forking/patching this for your use case - I will be merging `v2` to main branch soon

seunghyeoks:
> hey, I just released last version of `v1` if you don't mind forking/patching this for your use case - I will be merging `v2` to main branch soon

I'd be grateful if you could merge it. But does v2 have similar functionality? I haven't looked at v2 yet.

demchenkoalex:
I can merge that is no problem, just don't want to release since I already wrote that final version was released :)

`v2` is a complete redesign, it does not have proper UI yet, but solid/performant foundation only. But I think with the new foundation UI will be a quick thing. I will also focus on making everything modular, so missing parameters should not be a thing anymore.

Issue #182, opened 2022-01-03
Avatar should be shown first on multiple messages from the same user


itaishalom:
Hi, it is more common to display the avatar close to the first message of the user (if he sends multiple messages)
and not to the last as displayed today.
What do you think?

demchenkoalex:
Hey, sorry for the late response. Our thinking was that usually with the avatar name is enabled, and name is showed close to the top and avatar is on the bottom. So reading both ways you know the author. Don't think we will be changing that.

itaishalom:
Do you think that i could do something easily and maybe pr you?
Is it an easy change?

demchenkoalex:
Sure, you can make it an option, like a boolean or something :)

itaishalom:
From your perspective, is it a difficult change?

demchenkoalex:
Not sure, I would need to take a look

itaishalom:
Thanks

demchenkoalex:
Ok, so it is somewhere in the https://github.com/flyerhq/flutter_chat_ui/blob/f73eccb40b77c55297995c14a83bf71563a0920b/lib/src/util.dart#L89, specifically where we calculate all chat looks

```dart
if (showUserNames) {
      final previousMessage = isFirst ? null : messages[i + 1];

      final isFirstInGroup = notMyMessage &&
          ((message.author.id != previousMessage?.author.id) ||
              (messageHasCreatedAt &&
                  previousMessage?.createdAt != null &&
                  message.createdAt! - previousMessage!.createdAt! >
                      groupMessagesThreshold));

      if (isFirstInGroup) {
        shouldShowName = false;
        if (message.type == types.MessageType.text) {
          showName = true;
        } else {
          shouldShowName = true;
        }
      }

      if (message.type == types.MessageType.text && shouldShowName) {
        showName = true;
        shouldShowName = false;
      }
    }

    if (messageHasCreatedAt && nextMessageHasCreatedAt) {
      nextMessageDateThreshold =
          nextMessage!.createdAt! - message.createdAt! >= dateHeaderThreshold;

      nextMessageDifferentDay =
          DateTime.fromMillisecondsSinceEpoch(message.createdAt!).day !=
              DateTime.fromMillisecondsSinceEpoch(nextMessage.createdAt!).day;

      nextMessageInGroup = nextMessageSameAuthor &&
          nextMessage.createdAt! - message.createdAt! <= groupMessagesThreshold;
    }
```

[...]

PR #637, opened 2024-10-08
Check first message

berkeenesakt:
### What does it do?

It adds a new parameter to bubble builder named `isFirstInGroup`.

### Why is it needed?

It helps us to determine first message in group or a single message. I need it because in my situation I need to show user names and avatars on first message. To do it I need to check for first message but current version of chat_ui doesn't allow to check it. 
Screenshot 2024-10-08 at 07 27 19

### How to test it?

Change your flutter_chat_ui version on pubspec.yaml like that:
```
flutter_chat_ui:
    git:
      url: 'https://github.com/berkeenesakt/flutter_chat_ui.git'
      ref: 'feature-first-message-in-group'
```
Update your bubble builder like that
```
bubbleBuilder: (child, {required isFirstInGroup, required message, required nextMessageInGroup}) {
...
```
Now you can check for first message with `isFirstInGroup` parameter.

### Related issues/PRs

https://github.com/flyerhq/flutter_chat_ui/issues/182

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -136,17 +136,16 @@ List<Object> calculateChatMessages(
     var nextMessageInGroup = false;
     var showName = false;
 
-    if (showUserNames) {
-      final previousMessage = isFirst ? null : messages[i + 1];
+    final previousMessage = isFirst ? null : messages[i + 1];
 
-      final isFirstInGroup = notMyMessage &&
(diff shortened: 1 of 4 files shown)


demchenkoalex:
message grouping is not a thing anymore in `v2` - but I will try to make those values available for custom builders, like in this PR.

PR #641, opened 2024-10-14
add text_message_text_builder and previewImageBuilder

phamconganh:
### What does it do?

Add custom text message text builder. Example build markdown from flutter_markdown
Add previewImageBuilder for link preview


demchenkoalex:
part of `v2` by default

PR #689, opened 2024-12-24
Update chat.dart

Aakash574:
Hello, I'm Aakash Panwar I just add Background image to chat. can you please review and update this on next update.

please consider this 
thankyou

### DIFF
--- a/lib/src/widgets/chat.dart
+++ b/lib/src/widgets/chat.dart
@@ -55,7 +55,6 @@ class Chat extends StatefulWidget {
     this.emptyState,
     this.fileMessageBuilder,
     this.groupMessagesThreshold = 60000,
-    this.messagesSpacerHeight = 12,
     this.hideBackgroundOnEmojiMessages = true,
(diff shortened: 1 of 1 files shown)


lib/src/widgets/chat.dart
```diff
@@ -55,7 +55,6 @@ class Chat extends StatefulWidget {
     this.emptyState,
     this.fileMessageBuilder,
     this.groupMessagesThreshold = 60000,
-    this.messagesSpacerHeight = 12,
```
demchenkoalex, line 58: why you would remove this?

Aakash574: By mistake removed

demchenkoalex: are you changing this back?

Aakash574: Can you please add the backgroundImage in Package please. i don't want to remove any thing in package.

demchenkoalex: you already removed this, just bring it back :)

PR #712, opened 2025-02-14
fix Error: The non-abstract class 'CachedNetworkImage' is missing imp…

leeyisoft:
fix Error: The non-abstract class 'CachedNetworkImage' is missing implementations for these members: NetworkImage.webHtmlElementStrategy


demchenkoalex:
thanks for this! I had to bump flutter/dart to latest to be able to override this field, so I did all that on master. Closing this as it is already on master. Sorry about that.

PR #754, opened 2025-05-05
Fix insertion in reverse chat list always at position 0

nicolasbraun:
Hello @demchenkoalex 

In `ChatAnimatedListReversed` the insertion did not take the position into account, thus it was only working for new messages (at the bottom of the list),


nicolasbraun:
I opened a discussion and realize this breaks my PR so flagging this as draft for now

https://github.com/flyerhq/flutter_chat_ui/discussions/755

PR #756, opened 2025-05-06
Fix StreamOperation Async issues / create insertAll / fixInsert on reverted list

nicolasbraun:
Hello @demchenkoalex 

First off sorry for the big PR, stuff all depend on each other but I tried to make small commits for readability. 
This PR mostly aims at fixing the insertion bug from PR #754, that triggered the discussion #755 on the issues cause by the `OperationStream` being async (by nature) while the List directly relied on the chatController's `messages`.

### TODOs in code

The code still contains some TODOs, those are mostly suggestions / refactoring opportunities that need your approval

### PR breakdown

615a3de097734ea07370cb0187076e8cf28342f2 : Commit from #754 that aimed at fixing the insert always at index 0 in reverted List => this cause the crash discussed in #755
add8d88dd857c7f53cb751f85886f20b41d1beee to 98b22bd65fd4dc2bbfb184bc02f5aa475d2d4cdc  (and e262542d542b8cceac3ac0b48a7ceef665fa6ef5): Use the local state instead of directly relying on the controller's message => this can lead to conflicts between operation since they are executed in parallel
709818824a7a0af053d85c92930ab9ff4bcc16e6: Implement the queue to solve this
f45b94cc81020d5c1bcaa11ed4291f0da93a0f55: small refactor
d431ae3e08b8cc73ed2ddb1617ceea02148059b4: big factorization of the 2 widgets, more on this below
d51ee0bd8e528a839c6b06eeae81c893f11f531e: add the ability to insert a List of messages at once, without relying on diffutil for performance

### Factorization of the 2 widgets
I realized there was a lot of similar code between the reversed and normal list. 
It seems better for maintanability to factorize them; using some helpers for the small differences in behavior (mostly scroll).

I first started to write some mixins with abstract methods to override in each widget but this feeled way too complicated for 2 lists that were mostly the same.

Maybe we could still use mixins if we want to make the code more readable (something like `ChatListScrollManagementMixin, ChatListOperationsManagementMixin ChatListAnimationsMixins`)

Note I kept the `ChatAnimatedListReversed`widget for retrocompatibility and to keep the "bottom", "top" attributes properly named.

I you prefer i can easily remove this and duplicate code in the reversed List

### Tests

I tested on my app without issue. 
Will try to test on your example tomorrow


demchenkoalex:
I had 5 minutes to quickly go through the code - I’ll take a deeper look over the weekend. But yes, the two lists are quite similar, except for the scroll logic / pagination / maybe something else. I was also thinking about extracting the shared parts into mixins, but I was a bit lazy 😅 

If we can manage to merge them, that would be great. But I spent months testing each one for different use cases, and I really don’t want to go through all that again… though it seems like I might not have a choice 😄

nicolasbraun:
I feel you :)

I can help with testing, with my update I basically use ifs when the code
was different in the reversed list, so there should not be any regression.
I added the TODOs because sometimes I could not tell if this was an
omission of something you really wanted only in the non-reversed list (do
some people even used non reversed ? :p)

I did not dive into mixins because, for example, the scroll mixin would
have to have access to `_oldList`, this is doable with an "abstract" getter
in the mixin but (once again not a Dart pro) felt like not the right way to
do it.

Let me know if I can help, will have some time next week.

nicolasbraun:
I removed the update Operation management, I saw it's handled directly by each chatMessage

demchenkoalex:
Okay, it took me a solid 10 hours of work, but hopefully, it will benefit everyone. One thing led to another, and I ended up adding probably a thousand more lines, haha. Not sure if you want to review it @nicolasbraun or simply test it out.

Key Changes:

1. Integration Test for List Items:
    - Added integration tests that check all operations and validate the correct positioning of items after those operations.
2. `insertAllMessages` Added to `chatController`:
    - The operation was created earlier but wasn’t included in the controller - now it’s fully integrated.
3. Handling of `insertAllMessages` in Lists:
    - Implemented animation state for bulk insertions.
4. Controller Fixes for Latest Messages:
    - Now, the controller fetches the latest message before updating/removing. This was necessary because if an image’s dimensions change, the list couldn’t delete that message since the controller was sending the old version.
5. Proper Handling of `setMessages` in `chatController`:
    - Previously, setting messages would clear everything instead of respecting the provided array. Now, it correctly calculates the diff.
    - This led me to debug **DiffUtil** for at least 5 hours - during integration tests, I caught multiple “out of bounds” errors that others reported.
    - TL;DR: Since **DiffUtil** changes were applied sequentially, it often broke indices during complex diffs (especially during removals).
    - The solution was to add **move** support, which eventually fixed the crashes, as confirmed by two integration tests.
6. Significant Updates to **ChatAnimatedList**:
    - Fixed missing pieces during the merge.
    - Added more comments for clarity.

---

Answers to Your _TODOs_:
1. `_linkAnimationToScroll` for Reversed Lists:
    - Not done, because its purpose is to solve one specific issue: Flutter rarely scrolls to the exact end when going to `maxScrollExtent`. For reversed lists, we always scroll to 0, which is a known value and always works fine.
2. Multiple _TODOs_ in the Scrolling Section:
    - Solved by the point above, so I reinstalled the previous logic.
    - Where you asked why I didn’t just call scrollToBottom: I actually did that at first, but I remember facing issues. The current logic is more robust.
3. Reversed Lists and `_adjustInitialScrollPosition`:
    - I added the early return you suggested - reversed lists always start at 0, so this adjustment is unnecessary.
4. Set Operation Won’t Fallback to `_chatController.messages`:
    - If messages is null, it will perform a full clear instead - so it now falls back to [].

---

Issues Addressed:
1. Insert at Specific Index:
    - Fully tested with integration tests.
2. Bulk Insert (insertAllMessages) Support:
    - Added and handled with animations.
3. Async Issues and Operation Queue:
    - Thanks for the original idea 🫡

---

I am fine with merging this now, will wait for your approval 🙂

nicolasbraun:
Thanks @demchenkoalex , i'll give it a try tomorrow

nicolasbraun:
@demchenkoalex 

I tested on my code (which does not have all usecase's and looks fine !
Also looks at the changes from my code to yours, everything looks good especially the `insertAll` that will be smoother.

I had to look into other classes for my other PR, especially `ChatMessageInternal` and I see it also uses the chatcontroller's messages for `_resolveGroupStatus`, we have the same "async issue" so maybe it'd be more robust to rely on the Chatlist here too (that would have to be exposed).

As a side note i'm not familiar with `Provider`  but I'm wondering how those `context.read` work if you have several Chats (like Slack channels).

qeepcologne:
thank you so much for this great improvement.

For me it also make sense to add removeAll operation to the conroller.
currently we need to call removeMessage in a loop resulting in operations stream is called many times.
What do you think?

Usecase is select multiple messages via longpress and then delete them via burger menu (i think this is usual).

demchenkoalex:
Good idea, I see that animated list supports removeAll operation

demchenkoalex:
Related documentation articles:
1. [ChatController Explained](https://flyer.chat/docs/flutter/getting-started/architecture/#chatcontroller-explained)
2. [Persisted Controller](https://flyer.chat/docs/flutter/guides/persisted-controller/)

demchenkoalex:
> Good idea, I see that animated list supports removeAll operation

After almost finishing the feature, I checked the source code for `removeAllItems` - and for some reason, it just uses the same builder for all removed messages. So whatever you remove ends up being shown as the same message during the removal animation. That’s… pretty dumb 😄
```dart
  void removeAllItems(AnimatedRemovedItemBuilder builder, {Duration duration = _kDuration}) {
    for (int i = _itemsCount - 1; i >= 0; i--) {
      removeItem(i, builder, duration: duration);
    }
  }
```
So it’s safe to say we don’t have a proper removeAll for lists. Doing individual removals is also error-prone. We already have a well-tested approach: just call set with a new array that has the items removed, and **DiffUtil** will take care of the rest.

Will keep the source code here just in case this pops again in the future, but I doubt I will be adding this.

**chat_controller.dart**
```dart
  /// Removes multiple [messages] from the list.
  ///
  /// The [messages] parameter contains the messages to be removed, identified
  /// by their respective `id` values.
  ///
  /// - Use this for batch removals, like clearing multiple selected messages
  /// - The controller should emit a `ChatOperation.removeAll(messages)`
  ///   event to notify listeners about the batch removal.
  /// - When implementing, the `messages` parameter might contain outdated
  ///   instances if they have been updated elsewhere. To ensure correctness:
  ///   1. Identify each message in your data store using `message.id`.
  ///   2. Retrieve the *current* versions of those messages from your store.
  ///   3. The emitted operations must use these current versions.
  Future<void> removeAllMessages(List<Message> messages);
```

**chat_operation.dart**
```dart
  /// Creates a removeAll operation, for removing multiple messages.
  factory ChatOperation.removeAll(List<Message> messages) =>
      ChatOperation._(ChatOperationType.removeAll, messages: messages);
```

**hive_chat_controller.dart**
```dart
  @override
  Future<void> removeAllMessages(List<Message> messages) async {
    if (messages.isEmpty) return;

    final sortedMessages = List.from(this.messages);
    final messagesToRemove = <Message>[];
    final idsToDelete = <String>[];
    final idsToRemoveSet = messages.map((m) => m.id).toSet();

    for (final message in sortedMessages) {
      if (idsToRemoveSet.contains(message.id)) {
        messagesToRemove.add(message);
        idsToDelete.add(message.id);
      }
    }

    if (idsToDelete.isNotEmpty) {
      await _box.deleteAll(idsToDelete);
      _invalidateCache();
      _operationsController.add(ChatOperation.removeAll(messagesToRemove));
    }
  }
```

**in_memory_chat_controller.dart**
```dart
  @override
  Future<void> removeAllMessages(List<Message> messages) async {
    if (messages.isEmpty) return;

    final messagesToRemove = <Message>[];
    final idsToRemove = messages.map((m) => m.id).toSet();

    // Collect messages to remove (from end to start to avoid index issues)
    for (var i = _messages.length - 1; i >= 0; i--) {
      if (idsToRemove.contains(_messages[i].id)) {
        messagesToRemove.insert(0, _messages[i]);
        _messages.removeAt(i);
      }
    }

    if (messagesToRemove.isNotEmpty) {
      _operationsController.add(ChatOperation.removeAll(messagesToRemove));
    }
  }
```

**chat_animated_list.dart**
no implementation as at this point I released it does not make sense.

PR #766, opened 2025-05-09
Allow custom ImageProvider

nicolasbraun:
## Motivation

V1 allowed a custom `ImageProvider` to be defined with `imageProviderBuilder`.
This is a feature I need because I use the same cacheManager to display the chat's images (using [`CachedNetworkImage`](https://pub.dev/packages/cached_network_image) ( not the one from `crosscache`) in other part of the app.

Not sure I could do this with crossCache + it would mean losing cache on migration

### DIFF
--- a/packages/flyer_chat_image_message/lib/src/flyer_chat_image_message.dart
+++ b/packages/flyer_chat_image_message/lib/src/flyer_chat_image_message.dart
@@ -64,6 +64,9 @@ class FlyerChatImageMessage extends StatefulWidget {
   /// Position of the timestamp and status indicator relative to the image.
   final TimeAndStatusPosition timeAndStatusPosition;
 
+  /// ImageProvider, otherwise crossCache's CachedNetworkImage will be used.
+  final ImageProvider? imageProvider;
+
   /// Creates a widget to display an image message.
   const FlyerChatImageMessage({
     super.key,
@@ -82,6 +85,7 @@ class FlyerChatImageMessage extends StatefulWidget {
     this.showTime = true,
     this.showStatus = true,
     this.timeAndStatusPosition = TimeAndStatusPosition.end,
+    this.imageProvider,
   });
 
   @override
@@ -93,7 +97,7 @@ class FlyerChatImageMessage extends StatefulWidget {
 class _FlyerChatImageMessageState extends State<FlyerChatImageMessage>
     with TickerProviderStateMixin {
   late final ChatController _chatController;
-  late CachedNetworkImage _cachedNetworkImage;
+  late ImageProvider _imageProvider;
   late double _aspectRatio;
   ImageProvider? _placeholderProvider;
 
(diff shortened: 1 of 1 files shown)


nicolasbraun:
@demchenkoalex 

Wanted your thoughts on how to achieve this, this may impact the PR.

- I have another part of the app where the user can browser all images (querying messages with type ìmage`
- ideally i would like to have the functionality of the new `FlyerChatMessage` to display local data while uploading
-  I"m using Cloud Storage (through firebase) so I don't know the source url before upload (I don't have the `token`)

It works will with `CrossCache` with this flow

1. Add an Insert operation with an image message and source as local file path
2. Upload image, upload to repo
3. Listen for new messages and insert in chat controller.

I could declare my cache elsewhere and reuse it but (correct my if I'm wrong)
- I can only have one CrossCache when https://pub.dev/packages/cached_network_image allows several caches; I currently use this to have different caching (maxAge, item count) logic for chatImages vs profile picture for example
- CrossCache does not allow to define a max objects count, stale period 


I think I could use the `thumbHash` but since `source` is mandatory I still have to provide an URL.
Maybe we could make source optional during upload?

demchenkoalex:
**CrossCache** isn’t fully finished yet I agree. My idea was to make caching work on the web as well, which `cached_network_image` currently doesn’t support. That being said, `cached_network_image` is definitely more mature.

The reason I wanted each message type to be its own separate package was to allow people to override them with whatever they prefer - for example, using a different caching provider. It’s totally fine to just copy the whole message widget into your own codebase and drop **CrossCache** if that fits better.

Given this setup, do you think it would be easier if we added an option to pass a custom provider instead?

> I think I could use the thumbHash but since source is mandatory I still have to provide an URL.
Maybe we could make source optional during upload?

As for that part, I didn’t fully understand it. What does `thumbHash` have to do with caching?
And to make source optional during upload, would we need to make it optional for every instance? I’m not sure how else we could achieve that.

nicolasbraun:
Yeah sorry that was not very clear.

Replacing `CrossCache` with `cached_network_image` using my PR woks fine (as both are `ImageProvider` but this does not work to display the local file during upload as it does not support `data;...` sources.

My current flow is

1. locally add the message to the controller with `source:data....` 
2. It's visually displayed + overlay using the `UploadProgressMixin`
3. once properly upload I send it to the repo update the `source` to the proper remote URL
4. I get notified by the server and updates the `Message` 


Since I wanted to try to stick to `cached_network_image` I was thinking how to tweak this to have kinda the same behavior, and was thinking:
- In step 1 i could use a `thumb` to have the "local" image displayed (does not rely on `cached_network_image`)
But I have to set a source so this does not work (except maybe if i put as source a remote transparent image, which seemed a bit hacky)

nicolasbraun:
This is how I ended up doing it, seems to work well.

```dart
        imageMessageBuilder: (context, message, index) {
          return FlyerChatImageMessage(
              message: message,
              index: index,
              imageProvider: message.source.startsWith("http")
                  ? CachedNetworkImageProvider(
                      message.source,
                      cacheManager: ChatImagesCustomCacheManager.instance,
                    )
                  : null);
        },
```

demchenkoalex:
we still have to merge this right?

nicolasbraun:
Yeah!
Maybe we should add something in the docs too

nicolasbraun:
@demchenkoalex I see you made a release but (maybe) forgot this one :)

demchenkoalex:
I didn’t - this will go out as a minor update next. I had to release a major breaking change first. Right now, I’m working on a guide explaining how to create custom controllers, since people need to understand that there’s an internal list that must be kept in sync.

demchenkoalex (approved):

Issue #763, opened 2025-05-09
v2 / empy state background missing feature


qeepcologne:

demchenkoalex:
fixed by https://github.com/flyerhq/flutter_chat_ui/pull/771

PR #771, opened 2025-05-12
Add an empty list widget and builder

nicolasbraun:
## Motivation

Tentative to allow displaying a widget while the `chatList`  is empty as requested in #763 .
I defaulted to a centered text (same as v1)

I saw several options

- Replace the `sliverList` in `buildSlivers` : that seemed like the perfect one, as it allowed any Widget even another Sliver. But this caused a lot of issue since `_listKey.currentState` could not be defined on insert. It felt a bit too risky
- Add it as single element of the `SliverList` => hard to manage position, no sliver
- As an overlay in the Stack: current implementation. Looks quite ok except the centering does not take the `Composer` height into account

PS: build on top of #756 so to merge afterwards

### DIFF
--- a/packages/flutter_chat_core/lib/src/models/builders.dart
+++ b/packages/flutter_chat_core/lib/src/models/builders.dart
@@ -72,6 +72,9 @@ typedef ScrollToBottomBuilder =
 /// Signature for building the loading indicator shown when fetching more messages.
 typedef LoadMoreBuilder = Widget Function(BuildContext);
 
+/// Signature for building the empty chat list widget.
+typedef EmptyChatListBuilder = Widget Function(BuildContext);
+
 /// A collection of builder functions used to customize the UI components
 /// of the chat interface.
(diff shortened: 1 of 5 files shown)


nicolasbraun:
This does not work yet, the EmptyChatList is added to the stack but since the whole Stack does not rebuild on each Message, it stays there until it rebuilds...

demchenkoalex:
PR that this depends on is merged

qeepcologne:
i thought about a similar solution, but always adding the empty widget and turn it on/off via Visiblity

demchenkoalex:
LGTM :) just added some animation and fixed the positioning.

devberkay:
Somehow, emptyChatListBuilder doesnt allow clickable children to detect taps (gestures). I cant put any button inside this builder, taps can't be detected. @demchenkoalex

nicolasbraun:
Fixed in #787

PR #778, opened 2025-05-16
Fix update operation not persisted in the list

nicolasbraun:
Hello 

@demchenkoalex 
Sorry this PR complements #756 and sadly requires an API change.

### Motivation

Currently the list does not listen to `ChatOperation.update`. When such an operation happens, all `ChatMessages` listen to it and eventually react updating themselves.

This means if the widget is in memory it will properly update **but** 
- if it is discarded
- then rebuilt
it will use the value at it's index in `_oldList` and this value has not been updated so the value prior to the update will be displayed.

### Discussion

I modified the signature to pass the `index` because usually if a client performs an update he must have calcultated it, like in all your examples.
If we want retrocompatibility we could make it optional and search for the message of correct id in the _oldList but this seemed less efficient.

### Side notes
- I put a log in `ChatMessage` listener while debugging and weirdly on an update operation it was printed several times, not really sure why since I emitted only one event.

### DIFF
--- a/examples/flyer_chat/lib/hive_chat_controller.dart
+++ b/examples/flyer_chat/lib/hive_chat_controller.dart
@@ -63,7 +63,7 @@ class HiveChatController implements ChatController {
       _box.write(() {
         _box.putAt(index, newMessage.toJson());
         _operationsController.add(
-          ChatOperation.update(actualOldMessage, newMessage),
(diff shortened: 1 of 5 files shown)


demchenkoalex:
Hey! Good catch! I’ll merge this now and release a breaking update with all the list management changes we’ve worked on.

PR #784, opened 2025-05-20
Feat/link preview v2

### LINKED ISSUE #758: Implement linkPreview in v2
Hello @demchenkoalex 

I can dive a bit into it but first would like to discuss the best option to handle linkPreviews in V2.
I think we can rely on  the generic `chatMessageBuilder` and use the `bottomWidget`. 

We could then bascially redo what was done in V1 and use `flutter_link_previewer` to create the widget.
It could work as-is but `flutter_link_previewer` depends on `flutter_chat_types` so it's a bit of an useless -oudated- dependency.

Would you be ok to make `flutter_link_previewer` depends on chat-core, or, even better, standalone ?

_Note we have a type conflic between `LinkPreview` in chatCore and the `LinkPreview` Widget_

nicolasbraun:
### Motivation

Aims to fix #758 and allow easier implementation of linkPreviews in the chat.
This relies on the following PR https://github.com/flyerhq/flutter_link_previewer/pull/63  in `flutter_link_previewer` , I made it there for easier review, but once it's ok we can move it in this monorepo.

**LinkPreview**
- [x] Made `LinkPreview` depend on chatcore
- [x] Reworked the methods to return `null` if no preview since the new model `link` property is not optional
- [x] Added back the image information
- [x] Reworked the layout to allow to hide specifiy items, especially the `linkify` text which "repeats" the chatMessage text
- [x] ⚠️ Made `width` optional (default to infinity) and used a layoutBuilder to constrain the `_imageWidget`  

**ChatUi**
- [x] Added a builder for the `linkPreview`   
- [x] Adapted `FlyerChatMessage` and `SimpleMessage`   to display the Preview
- [ ] No default implementation  => did not want to make them depends on `flutter_link_previewer` + the logic to fetch or not preview is to be decided by the user.

### Implementation example

```dart
linkPreviewBuilder: (context, message) {
          final metadataLinkPreviewFetchedKey = 'linkPreviewDataFetched';
          final isLinkPreviewDataFetched =
              message.metadata?.containsKey(metadataLinkPreviewFetchedKey) ??
                  false;

          if (message.linkPreviewData != null || !isLinkPreviewDataFetched) {
            return Column(
              children: [
                Divider(
                  height: 0.5,
                  color: Colors.white,
                ),
                LinkPreview(
                  text: message.text,
                  linkPreviewData: message.linkPreviewData,
                  hideDescription: false,
                  hideText: false,
                  hideTitle: false,
                  linkStyle: TextStyle().copyWith(color: Colors.white),
                  openOnPreviewImageTap: true,
                  openOnPreviewTitleTap: true,
                  onLinkPreviewDataFetched: (linkPreviewData) {
                    _controller?.updateMessage(
                      message,
                      message.copyWith(
                        metadata: {
                          ...message.metadata ?? {},
                          metadataLinkPreviewFetchedKey: true,
                        },
                        linkPreviewData: linkPreviewData,
                      ),
                    );
                  },
                ),
              ],
            );
          }
          return null;
        },
     
```

### Questions / notes

1. Not sure why the display as row is conditionned to the image being square? Maybe we could add an option to force a row/colum render
2. Understanding the layout was quite harsh, I tried to make it a little bit more readable

### Screenshots

| Preview | Description |
|--------|-------------|
|  | Everything showed |
|  | Hidden text |
|  | Title and image only |
|  | No img |
|  | img only |
|  | fake 1 ratio => row render |


packages/flyer_chat_text_message/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_chat_core: ^2.2.1
+  flutter_chat_ui: ^2.2.1
```
demchenkoalex, line 16: cycle dependency - we can't do this. looks like link preview must be in a separate package.

demchenkoalex: ah looks like we only need link preview position, so it should be moved to core package to solve this

demchenkoalex:
some changes I made after merging: https://github.com/flyerhq/flutter_chat_ui/commit/c922a0ca3879baf9b29387cbb2a74690aa1061f8

packages/flyer_chat_text_message/pubspec.yaml
demchenkoalex: Not technically. I guess it was circular in my head, when I had them as dependencies for flutter_chat_ui long time ago. In any case, the idea is that chat ui consumes those widgets and I don't want them to know anything about the chat ui - logically they should not care if they are in the chat ui, or maybe even in some other ui 🤷

PR #788, opened 2025-05-23
Add linkPreview example

nicolasbraun:
Figured i could push my test as example

The dependency still has to be updated once release, flagging as draft for now


nicolasbraun:
Closing in favor of #790

PR #789, opened 2025-05-23
Feat/reactions

nicolasbraun:
# Motivation

This is a first - partial - attemp to add reactions support to FlyerChat.
I raises a few architecture questions to discuss.

# Implementation
Currently the implementation is mostly based on 3 widgets
- `ReactionTile` represent a small bubble with the emoji and count
- `FlyerChatReactions` lays them out based on the available space + size calculations
- `ReactionList` displays all the reactions

I made my best to avoid using `InstrinsicSizes` for performance

# Questions

1. I opted for the Signal/Whatsapp style (and not something like Telegram)
2. Need to decide how to trigger reactions, since `onLongPressed` is already exposed at the `Chat` level

#  Know issues

1. Currently each `ReactionTile` handles it's internal state but this means the parent `FlyerChatReactions` can't eventually recalculate position to hide/display a new reaction
3. Needs a better way to manage the reaction height

# TODO

**Arch**
- [ ] Add a `reactionBuilder`: should it be a parameter of `ChatMessage` or Chat?; should `ChatMessage` use one by default?

**Ui**

- [ ] Cleanup customization options
- [ ] Make reactions list sexier
- [ ] Insertion animations? That would be for you alex :) Relates to the know issue above, maybe the tile should notify a count change and the Parent widget should redraw with style. I'm not used to flutter animations / list inserts

**Feat**

- [ ] Implement add reaction. I was thinking something like https://pub.dev/packages/flutter_chat_reactions is quite smart. Maybe not use the package but do something similar.

https://github.com/user-attachments/assets/2b76f77d-31a8-4815-a347-182f062a1b6f

### DIFF
--- a/examples/flyer_chat/lib/pagination.dart
+++ b/examples/flyer_chat/lib/pagination.dart
@@ -2,6 +2,7 @@ import 'dart:math';
 import 'package:flutter/material.dart';
 import 'package:flutter_chat_core/flutter_chat_core.dart';
 import 'package:flutter_chat_ui/flutter_chat_ui.dart';
+import 'package:flyer_chat_reactions/flyer_chat_reactions.dart';
 import 'package:provider/provider.dart';
 
 import 'widgets/composer_action_bar.dart';
@@ -21,7 +22,7 @@ class PaginationState extends State<Pagination> {
           final numLines = random.nextInt(4) + 1;
           final text = List.generate(
             numLines,
-            (lineIndex) => 'Message ${i + 1} - Line ${lineIndex + 1}',
+            (lineIndex) => 'Message  ${i + 1} - Line ${lineIndex + 1}',
           ).join('\n');
           return Message.text(
             id: (i + 1).toString(),
@@ -31,10 +32,17 @@ class PaginationState extends State<Pagination> {
               isUtc: true,
             ),
             text: text,
+            reactions: {
+              '👍': ['me'],
+              '👎': ['user2', 'me'],
+              '🥨': ['author'],
+              '👌': ['me', 'user3', 'user4'],
+              '👊': ['me'],
+            },
           );
         }).reversed.toList(),
   );
-  final _currentUser = const User(id: 'me');
+  final _currentUser = const User(id: 'me', name: 'This is me');
 
   MessageID? _lastMessageId;
   bool _hasMore = true;
@@ -46,6 +54,10 @@ class PaginationState extends State<Pagination> {
     super.dispose();
   }
 
+  void _onMessageReaction(int index, String? reaction) {
+    print('reaction: $reaction');
+  }
+
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
@@ -53,6 +65,7 @@ class PaginationState extends State<Pagination> {
     return Scaffold(
       appBar: AppBar(title: const Text('Pagination')),
       body: Chat(
+        onMessageReaction: _onMessageReaction,
         builders: Builders(
           chatAnimatedListBuilder: (context, itemBuilder) {
             return ChatAnimatedList(
@@ -179,6 +192,13 @@ class MockDatabase {
         isUtc: true,
       ),
       text: text,
+      reactions: {
+        '👍': ['me'],
+        '👎': ['me'],
+        '🥨': ['author'],
+        '👌': ['me'],
+        '👊': ['me'],
+      },
     );
   });
 
--- a/examples/flyer_chat/pubspec.yaml
+++ b/examples/flyer_chat/pubspec.yaml
@@ -2,7 +2,7 @@ name: flyer_chat
 description: "A new Flutter project."
 # The following line prevents the package from being accidentally published to
 # pub.dev using `flutter pub publish`. This is preferred for private packages.
-publish_to: 'none' # Remove this line if you wish to publish to pub.dev
+publish_to: "none" # Remove this line if you wish to publish to pub.dev
 
 # The following defines the version and build number for your application.
 # A version number is three numbers separated by dots, like 1.2.43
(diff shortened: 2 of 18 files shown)


demchenkoalex:
You're on fire 🚀 

How urgent is this? I was planning to wrap up some smaller issues this weekend, and next week I’ll be away at a conference. Realistically, I could spend some time on it next weekend.

nicolasbraun:
I was cooking this one for a while, did not have time to properly implement it.
No worries, no rush it was mostly to start a discussion

nicolasbraun:
Just to keep up. I made a quick test with the HeroTransition pattern and this raises quite a few issues

- Since we push a new route we loose `context, hence the chatBubble we are trying to rebuild does not have access to thr `Provider` stuff except if my re-provide them

```dart
// Example in ChatMessage
            onLongPress: () {
              final theme = context.read<ChatTheme>();
              final builder = context.read<Builders>();
              final user = context.read<UserID>();
              final timeFormat = context.read<DateFormat>();

              Navigator.of(context).push(
                HeroDialogRoute(
                  builder: (context) {
                    return MultiProvider(
                      providers: [
                        Provider.value(value: theme),
                        Provider.value(value: builder),
                        Provider.value(value: user),
                        Provider.value(value: timeFormat),
                      ],
                      child: Builder(
                        builder: (context) {
                          return ReactionsDialogWidget(
                            id: message.id,
                            messageWidget: ChatMessageWidget(
                              message: message,
                              isSentByMe: isSentByMe,
                              showReactions: false,
                              child: child,
                            ),
```

This is a bit tedious, and prone to error because we must be sure to pass down all the Provider that might be needed in the tree (by `SimpleTextMessage`,  `FlyerChatTextMessage` or any custom implementation (but they should not rely on our Provider))

- I think we also need to extract `_buildMessage` (if done in `ChatMessage`) to be a proper widget because it has to build using the proper `BuildContext` (cf code above)

- If we use a HeroTransition Flutter we might run into context-issue anyway if the "heroed" widget depends on Provider (Flutter tries to reuse it but if for some reason it rebuilds ... it may break)


Out of my head I see only those solutions

1. Dont use Hero, simply pass down providers and display a nice page with the chatBubble in the middle, but we loose the `HeroAnimation` and as seen above it makes a lot of "unsafe" code
2. Don't navigate: somehow display a blurry Widget on top of the `ChatList` Stack, without transition on the Bubble
3. Move the Multiprovider around the top Navigator => not an option there are so many options would be awkward
4. Use a non context-bounded state management system like `Riverpod`. It's built on top of Provider. This ensure our provider are available no matter the context. Though the app (or at least our Lib) must be inside a `ProviderScope`. Not sure how i behaves if the user already uses riverpod ( I asked here https://github.com/rrousselGit/riverpod/discussions/2569) and can test on my side


Before going any further lets maybe decide of the architecture, (on which widget we implement, which builder, declared where etc ..)

demchenkoalex:
Oh man, that’s a lot of problems 😅  I’m actually a React Native engineer - basically all my Flutter experience comes from building this library over the years. So whenever navigation is involved, I just pray to Cursor 😄

I’ll probably need to code the MVP myself from scratch to get a better understanding of what we should do.

nicolasbraun:
Yeah, well it's mainly one "big" problem => if we navigate we loose context so elements in the new tree do not have easy access to providers.

The other topics are mostly architectural to decide were to put the builders declarations, if we have defaults ones

I'll commit my code with many comment so you can eventually look when you got time.
Also i'll make a quick test with riverpod to see how it behave when nesting providers

nicolasbraun:
OK @demchenkoalex  I played a bit with `[Hero](https://docs.flutter.dev/ui/animations/hero-animations)` (like you I'm new to Flutter) and can explain a little bit more what happens, nothing we can't fix I think.

You can see base  (dirty!) code with comment on commit 77b5d7b4d5b835c405ca3b4c55c943fdbc958e5b

During a Hero transition, Flutter animates the same widget from page1 (context =  `fromHeroContext`) to page2 (context =  `toHeroContext`) and the animated widget is also build durinf transition in a thirdContext (let's call it `transitionContext`).

The problem we have is always the same: our widget use `Provider` in `context`, so we need to pass them. Like I sait there are not many solutions:

**1. Pass down context**
 Pass them down to the `toHeroContext` (and eventually to the `transitionContext`, see comment [here](https://github.com/nicolasbraun/flutter_chat_ui/blob/77b5d7b4d5b835c405ca3b4c55c943fdbc958e5b/packages/flutter_chat_ui/lib/src/chat_message/chat_message.dart#L138 , you can see example with a colored box in the video below) for the widget to be able to resolve.

This is prone to error but we could build a helper wrapper to be sure we don't forget to add some if whe add them.
Something like

```dart
class ChatMultiProviderWrapper extends StatelessWidget {
  const ChatMultiProviderWrapper({super.key, required this.child});
  final Widget child;

  @override
  Widget build(BuildContext context) {
    final theme = context.read<ChatTheme>();
    final builder = context.read<Builders>();
    final user = context.read<UserID>();
    final timeFormat = context.read<DateFormat>();
    // Read all providers, with optional if needed

    return MultiProvider(
      providers: [
        Provider.value(value: theme),
        Provider.value(value: builder),
        Provider.value(value: user),
        Provider.value(value: timeFormat),
        /// etc ..
      ],
      child: child,
    );
  }
}
```

https://github.com/user-attachments/assets/ef9d7ef0-05f7-4a55-bb25-1f514c6dac40


**2. Use a non-context based state management (leaning to riverPod that i know but there might be others**

This way it's "easy" the widget can build wherever they are.
I made a quick demo here 0df0ac26a6d3ea88b94811a9c7b434e0a8f1a5ea

Note it's a bit messy because I still have Provider  (only did the theme)
Also I did not declare the `providerScope` in the lib yet (we need to wrap eventually to avoid user to do it).



https://github.com/user-attachments/assets/39764145-16a9-4ec2-8ac7-38534df7d07a



**3 Don't use a hero transition**
- add reaction another way: could be a `bottomSheet`, a widget that appears in the `animatedList` Stack but its not state of the art.

nicolasbraun:
@demchenkoalex 

I'll make a last big push on this one, but open a new clean PR and close this one in favor of the new one, to remove all the tests/iterations made along this one.

I'll explain in the new one the problems I faced and how i went around them + the "architectural" points that may remain.

PR #790, opened 2025-05-23
Feat/link preview package

nicolasbraun:
As discused in the former link_preview repo https://github.com/flyerhq/flutter_link_previewer/pull/63 this PR aims at

- moving the preview as a package in this monorepo
- removing dependency to `flutter_chat_types`
- adapting the style to fit what messaging apps usually do

# Sidenote

@demchenkoalex 
I had to rewrite the `SimpleTextMessage`and `FlyerTextMessage` to account for both the timeAndStatus and the previex position.

Also i noticed in your Stack you had an hidden timeAndStatus
https://github.com/flyerhq/flutter_chat_ui/blob/a0c5eca94e47ec42acd31d10a25ac4657d0e1b6b/packages/flutter_chat_ui/lib/src/simple_text_message.dart#L190
I suspect it was to make the bubble big enough for the time but this has no effect if not put inside a column (it stacks on the textContent).
I  think it actually worked because of this line

https://github.com/flyerhq/flutter_chat_ui/blob/a0c5eca94e47ec42acd31d10a25ac4657d0e1b6b/packages/flutter_chat_ui/lib/src/simple_text_message.dart#L187

I used the same trick

```dart
            if (timeAndStatusPosition == TimeAndStatusPosition.end)
              SizedBox(height: textStyle?.lineHeight ?? 0),
```

but if you prefer we can add a hidden `timeAndStatus` so we are sure it's the proper size

# Screens

Simulator Screenshot - iPhone 16 Plus - 2025-05-23 at 19 23 21
Simulator Screenshot - iPhone 16 Plus - 2025-05-23 at 19 23 18
Simulator Screenshot - iPhone 16 Plus - 2025-05-23 at 19 23 14


nicolasbraun:
PS : maybe we can could expose the message backGroundColor and/or the isSentbyMe in the builder to adapt colors to the chatMessage.

I did not want to make this package use `Provider` and recalculate colors based on `ChatTheme` to make it standalone but we could

nicolasbraun:
**MacOS with width constrain of MediaQuery 10%**

Capture d’écran 2025-05-24 à 09 11 35

nicolasbraun:
Fixes + rebase

demchenkoalex:
had to re-write the widget from scratch as I didn't like that it was stretching for the whole width, so I added support of min and max width of the preview itself, using calculation of the parent text. From my testing everything is working nicely! I also changed package name to the original one to keep the same project on pub.dev

PR #794, opened 2025-05-28
Allow to set a different color in Composer when not empty

nicolasbraun:


demchenkoalex:
```dart
onChanged: (value) {
  setState(() {});
},
```

will hurt performance, I will use `ValueNotifier` instead.

demchenkoalex:
also renamed param to `emptyFieldSendIconColor` I feel like setting color for **empty** state is more straightforward

PR #798, opened 2025-06-02
Feat/model update Fix #797

### LINKED ISSUE #797: Pinned Message discussion
Hello @demchenkoalex 

### Model

Would you be open to add a `pinnedAt` (and maybe `pinnedBy`) field on the model ? (or just `pinned`)

This seems like a standard feature, and `Message` being sealed it's quite complicated to add fields without a lot of code or "unsafe" code reading metadatas

### Best way to have a sticky/floating header on reversed list

Since reversed list "start" is at the bottom sadly using a `SliverPersistantHeader` (more specificaly `SliverFloatingHeader`)  as `topWidget` does not work to have a nice sliver that dissapears when scrolling top, appears when scroling a bit down.

Here the intended behavior using a non-reversed list , and how it behaves with a reversed one (make sense, those widgets are not made for this, not sure why it flashes though, on the dartpad demo demo it goes under the list).

https://github.com/user-attachments/assets/143e33d9-ad05-4927-a336-8f9d63300ced

https://github.com/user-attachments/assets/3927484f-3ef0-44d4-8aad-b99c75cf41c7

**Solutions**

Currently I simply use a Column on top of the chatList, this works for my pinned messages, but I loose any possibility to hide/show on scroll, maybe we could expose the scrollNotifications(very low priority).

Could not find any better solution but happy to hear any idea. (don't want to use non-reversed list because of the performance issue we discussed in other issues + i like message being sticky to the bottom when the list is short, and this is quite hard to achieve with the non reversed list (i think it would required making it a non-sliver list and use `SliverFillRemaining`))

**Sidenote on Composer**

I was wondering why you used a Stack for the composer ? Seems more pratical to use a Column what would automatically anchor it at the bottom of the list, was it to be able to position it anywhere  (not sure there is a usecase). Wonder if it would make the keyboa//rd handling / code easier on this part 

I'm asking because this prevents bottom widgets to be properly sticky, has the scrollview goes behind the composer.
Example videos with a dummer `SliverPersistentHeader` (second one as the Composer offseted to see)

https://github.com/user-attachments/assets/a8da84c9-d699-4a9d-822c-0abe340d0797

https://github.com/user-attachments/assets/309fe765-f447-4cd1-8ccb-bd15f1432ada

nicolasbraun:
As discussed in #797 

- Add `pinned` bool field on `Message` 

Took the opportunity to add `editedAt`, `editedByUserId` on `TextMessage`  which is also a classic usecase to display when a message as been edited (maybe it should be on `Message`). `updatedAt` is not enough because a message can be updated for many reasons (like adding previewData).

If you are ok with those change I could propose an update to `FlyerChatTextMessage` and `SimpleTextMessage` to display to "Modified" mention.

Also I'm wondering if 
-a `deletedByUserId` field would by a classic usecase too. (In many apps admins can delete other people messages).
- text should no be on all models, many apps allow to send attachments with a text

### DIFF
--- a/packages/flutter_chat_core/lib/src/models/message.dart
+++ b/packages/flutter_chat_core/lib/src/models/message.dart
@@ -46,6 +46,12 @@ sealed class Message with _$Message {
     /// Timestamp when the message was last updated.
     @EpochDateTimeConverter() DateTime? updatedAt,
 
+    /// Timestamp when the message was last edited.
+    @EpochDateTimeConverter() DateTime? editedAt,
+
+    /// ID of the user who last edited the message.
+    UserID? editedByUserId,
+
     /// Map of reaction keys to lists of user IDs who reacted.
     Map<String, List<UserID>>? reactions,
 
--- a/packages/flutter_chat_core/lib/src/models/message.dart
+++ b/packages/flutter_chat_core/lib/src/models/message.dart
@@ -49,6 +49,9 @@ sealed class Message with _$Message {
     /// Map of reaction keys to lists of user IDs who reacted.
     Map<String, List<UserID>>? reactions,
 
+    /// Indicates if the message is pinned.
+    @Default(false) bool pinned,
+
     /// Additional custom metadata associated with the message.
     Map<String, dynamic>? metadata,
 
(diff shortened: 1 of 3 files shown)


nicolasbraun:
@demchenkoalex Done 🚀

demchenkoalex (approved):

Issue #750, opened 2025-05-03
Please provide Firebase documentation to support migration to V 2.0!


philipmjohnson:
Congratulations on releasing Version 2.0.  It was clearly a lot of work and the new architecture is a great improvement.

As a 1.0 user, my biggest obstacle to migration is figuring out how to either (a) use my existing firebase collections with the new version, or (b) migrate the existing collections to a new structure that is more appropriate for 2.0. 

It would be really useful to update the documentation in two ways:

1. Provide an example Firebase backend implementation for 2.0 (this assumes a "fresh" database).  That would help us Firebase users understand what the appropriate "goal state" for the Firebase collections would be.

2. Provide some assistance in the migration guide for those of us migrating from 1.0 and using Firebase (which, I assume, is basically everyone using 1.0). 

Thank you so much for your support of the Flutter community with this awesome package.

nicolasbraun:
Hello @philipmjohnson 

I'm currently in the process of migrating so can give you a few hints.

**1. Backend implementation**

The implementation of the firebase implementation in V1 was done in `flutter_firebase_chat_core` so depending on what you already have the [previous implementation](https://github.com/flyerhq/flutter_firebase_chat_core/blob/main/lib/src/firebase_chat_core.dart) is a very solid base for interactions with firebase.

 You might have to drop some other keys when writing a message (`seenAt`... should not be stored in the repository) but the base is here.

If thats of any help here is my current mapper to create a map from a `Message` that is then used to create the firebase doc .

> [!WARNING]  
> It's WIP I did not finish migrating yet

```dart
import 'package:cloud_firestore/cloud_firestore.dart';
import 'package:flutter_chat_core/flutter_chat_core.dart' as chatcore;
import 'package:flutter/foundation.dart' show immutable;
import 'package:moodapp/utils/log.dart';

class ChatsFireStoreMapper {
  static Map<String, dynamic> toFirestoreMap(
    chatcore.Message message,
  ) {
    try {
      var map = message.toJson();
      map[ChatMessageKey.updatedAt] = FieldValue.serverTimestamp();
      final dateFields = [
        ChatMessageKey.createdAt,
        ChatMessageKey.deletedAt,
      ];

      // Convert date fields to Firestore timestamps
      for (final field in dateFields) {
        if (map[field] != null) {
          map[field] = Timestamp.fromDate(DateTime.parse(map[field]));
        }
      }

      // Remove status-related fields that are not stored in Firestore
[...]

PR #799, opened 2025-06-02
add UserID? peerId ; add MessageType

leeyisoft:
add a peerId of the same level as authorId for all types of messages. It is defined as follows:

```
/// ID of the other party. If it's a one-on-one chat, it's the friend's ID. If it's a group chat, it's "GroupId". If you are chatting with a robot, he is botId
@override final  UserID? peerId;
```

add MessageType copy from ./flutter_chat_types-main/lib/src/message.dart and add textStream
```
/// All possible message types.
enum MessageType {
  audio,
  custom,
  file,
  image,
  system,
  text,
  unsupported,
  video,
  textStream,
}
```

### DIFF
--- a/packages/flutter_chat_core/lib/src/models/message.dart
+++ b/packages/flutter_chat_core/lib/src/models/message.dart
@@ -167,6 +167,9 @@ sealed class Message with _$Message {
     /// Height of the image in pixels.
     double? height,
 
+    /// Size of the image in bytes.
+    num? size,
+
     /// Indicates if an overlay should be shown (e.g., for NSFW content).
     bool? hasOverlay,
   }) = ImageMessage;
--- a/packages/flutter_chat_core/lib/src/models/message.dart
+++ b/packages/flutter_chat_core/lib/src/models/message.dart
@@ -22,6 +22,9 @@ sealed class Message with _$Message {
     /// ID of the user who sent the message.
     required UserID authorId,
 
+    /// The unique identifier of the other party. If it's a one-on-one chat, it's the friend's ID. If it's a group chat, it's "GroupId". If you are chatting with a robot, he is botId
+    UserID? peerId,
+
     /// ID of the message this one is replying to.
     MessageID? replyToMessageId,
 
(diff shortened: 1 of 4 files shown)


nicolasbraun:
For reference this is the following of a discussion in # 750
https://github.com/flyerhq/flutter_chat_ui/issues/750#issuecomment-2929619993


**PeerId**
Can you elaborate on the `peerId` needs ? It seems to me it's more related to a `room` than to each `Message`. 
V2 does not manage rooms at all so we are free to do whatever we need.

**Status enum**
Are you switching on the `type` field of the JSON serizalization directly ?
`Message` does not really have a type anymore but you can switch in it's `Type` 

```dart
switch (message) {
  case TextMessage():
    return XXX;
  case ImageMessage():
    return YYY;
  // ... other cases
  default:
    return ZZZ
}
``` 

I understand in a Firebase => Model mapper you may want to have some logic related to the type.
We could add the `enum` but we would have to keep in mind the value in the Json serialization would not come from it using the current codebase (it comes  directly from the factories names for example `textStream` [here](https://github.com/flyerhq/flutter_chat_ui/blob/80d53375449bb8300281c967f2046c63c980d5c1/packages/flutter_chat_core/lib/src/models/message.dart#L64))

leeyisoft:
> **PeerId**
> Can you elaborate on the `peerId` needs ? It seems to me it's more related to a `room` than to each `Message`.
> V2 does not manage rooms at all so we are free to do whatever we need.

My understanding:

Any message involves two parties or N+1f parties (N>=1). If there is an authorId in the model, there should be a peerId.

If there is this peerId, it can also be understood as the roomId of the chat room. Wouldn't this be more flexible?

demchenkoalex:
This is a Firebase-specific field and will never be added to `v2`.

It’s important to understand the separation between frontend (UI) and backend (Firebase, REST, etc.). The UI doesn’t care about things like peerID, remoteID, or roomID - its only job is to display messages. It doesn’t need to know where they came from.

`v1` blurred that separation, which was a fundamental architectural mistake.

In this case, if I remember correctly, peerID isn’t even used within the chat itself - it’s used on the list of chats, where you display the avatar of the person you’re talking to. So adding it to the message is pointless from a UI perspective - it has no effect on how the message is shown.

Example:
User A sends a message “Hello!” to user B.
	•	authorId is userA,
	•	The UI shows “Hello!” with user A’s name and avatar (resolved by authorId).
	•	You could set peerID = userB, but the UI doesn’t use that at all.

If a field doesn’t affect the UI, then it doesn’t belong in the UI layer.

Regarding the added type enum - this has been replaced in `v2` by union types, so it’s now derived from the message class itself. You won’t be able to use or compare to that field anyway.

PR #801, opened 2025-06-02
Feat/video message

nicolasbraun:
### Implementation notes
- I do not rely on Chewie's player for the bubble because I did not manage to have only the play button and entering fullScreen  with chewie's can break device orientation
- I did not allow to go "real fullscreen" (what will make the app eventually go landscape) because it adds a lot of complexity but we could look into it.

Maybe remaining points
- [ ] Caching, not sure how and if we should for videos
- [x] Maybe videoMessage could have a thumbnail but we can do this later
- [x] I'm not sure it works with a local video, especially on web
- [ ] Note the tap overrides the one from `ChatMessage` (at least on the bubble). But if we want to prevent that the user would have to handle video playback.

https://github.com/user-attachments/assets/668b0764-738f-4a78-85aa-b73966dc721e

### DIFF
--- a/packages/flyer_chat_video_message/pubspec.yaml
+++ b/packages/flyer_chat_video_message/pubspec.yaml
@@ -10,12 +10,13 @@ environment:
   flutter: ">=3.29.0"
 
 dependencies:
-  chewie: ^1.11.3
+  chewie: ^1.12.1
   flutter:
     sdk: flutter
-  flutter_chat_core: ^2.4.0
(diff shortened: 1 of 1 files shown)


Henry-Hiles:
Is this ready to use? I might just take a couple of your patches onto a fork of mine to use, if that's okay?

nicolasbraun:
PS: (i may need some adjustement if the main branch moved a lot, did not pull in a while)

PR #817, opened 2025-06-12
add method OnMessageDoubleTapCallback; add param BuildContext context for OnMessageTapCallback and OnMessageLongPressCallback

leeyisoft:
add method OnMessageDoubleTapCallback;
add param BuildContext context for OnMessageTapCallback and OnMessageLongPressCallback

The following functions must be in the context of the message widget
```
  // 消息长按事件
  void _onMessageLongPress(
      BuildContext c1,
    Message message, {
    int? index,
    LongPressStartDetails? details,
  }) {
    iPrint('_onMessageLongPress');
    // BuildContext c1 = getx.Get.context!;
    final menu = popupmenu.PopupMenu(
      context: c1,
      items: logic.getPopupMenuItems(message),
      onClickMenu: onClickMenu,
    );

    final renderBox = c1.findRenderObject() as RenderBox;
    final offset = renderBox.localToGlobal(Offset.zero);

    double l = offset.dx / 2 - renderBox.size.width / 2 + 75.0;
    double r = renderBox.size.width / 2 - 75.0;
    double dx = message.authorId == UserRepoLocal.to.currentUid ? r : l;
    double dy = offset.dy.clamp(0, getx.Get.height);
    double h = renderBox.size.height > getx.Get.height
        ? getx.Get.height
        : renderBox.size.height;

    menu.show(rect: Rect.fromLTWH(dx, dy, renderBox.size.width, h));
  }

```

### DIFF
--- a/packages/flutter_chat_ui/lib/src/chat_message/chat_message.dart
+++ b/packages/flutter_chat_ui/lib/src/chat_message/chat_message.dart
@@ -123,6 +123,7 @@ class ChatMessage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final onMessageTap = context.read<OnMessageTapCallback?>();
+    final onMessageDoubleTap = context.read<OnMessageDoubleTapCallback?>();
     final onMessageLongPress = context.read<OnMessageLongPressCallback?>();
     final isSentByMe = context.read<UserID>() == message.authorId;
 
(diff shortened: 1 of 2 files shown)


nicolasbraun:
@demchenkoalex i did not try, so maybe it works if @leeyisoft did but i think we missed providing this to provider here

https://github.com/flyerhq/flutter_chat_ui/blob/79672bfa584a8f4c46369570eb4f4bb8a87fc83a/packages/flutter_chat_ui/lib/src/chat.dart#L137

(On a side note i think Provider relies on "primitive" types, even if my use typedef for our callback.
We might run on some problem if two callback end up having the same signature (or if we have another String provider like the UserID one)

demchenkoalex:
I had many reports saying I should ditch **Provider**, so wanted to look into that after I merge your PRs @nicolasbraun

PR #832, opened 2025-07-01
Fix composer send color not reset on send

nicolasbraun:


demchenkoalex:
I do this inside the if (so when the message is sent) - not sure how this fix helps?

demchenkoalex:
Screenshot 2025-07-03 at 18 56 02

sorry I don't get it still - it is called 2 lines above? if it is not in that condition, then message is not sent anyway

nicolasbraun:
My bad between the time i fixed it it my local codebase and the time i made the PR a new commit was added.

7a496607966f0e976ccefdc80ac5b42e4bf59f8f

Thanks alex

PR #847, opened 2025-08-18
Support --wasm builds by adding dart.library.js_interop case to cross_cache

dh0er:
dart.library.html is not available for WASM builds, so cross cache currently fails with

`'Cache is not available in your current platform.'`

Since idb_shim is WASM compatible and doesn't rely on html, we can just add dart.library.js_interop to select the web implementation.


i11010520:
Hi @dh0er, when will this PR be available?

dh0er:
That's not for me to decide, I'm not a maintainer and also waiting for it to get merged.
If you don't want to wait you can use my fork. Simply add this to your pubspec.yaml:

dependency_overrides:
  cross_cache:
    git:
      url: https://github.com/dh0er/flutter_chat_ui.git
      path: packages/cross_cache
      ref: wasm_support

PR #858, opened 2025-10-06
(feat): fix link adding underline with color

GrandjeanG:
Currently links are not taking the color link, I added this PR so that we can still add a text decoration with specific color to the links. 
I was not able to change the text color, even with many different techniques.

### DIFF
--- a/examples/flyer_chat/.gitignore
+++ b/examples/flyer_chat/.gitignore
@@ -27,11 +27,11 @@ migrate_working_dir/
 **/doc/api/
 **/ios/Flutter/.last_build_id
 .dart_tool/
-.flutter-plugins
 .flutter-plugins-dependencies
 .pub-cache/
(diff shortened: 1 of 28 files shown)


TJMusiitwa:
@demchenkoalex 
It would be great, if we could get this fix launched as for some reason on the version I am using, even setting the links color does not seem to modify it 
image

The link is there but a user can't tell that it is there.

demchenkoalex:
sorry about that, deploying now

demchenkoalex:
guys this is not fixed by this PR 😅 this PR adds decoration, but you still can't set link color - [it is a bug in gpt_markdown](https://github.com/Infinitix-LLC/gpt_markdown/issues/85) - no matter what I do I can't set color, unless I do some recursion or create completely new span which will drop formatting.

PR #864, opened 2025-11-11
various improvements

Henry-Hiles:
Adds some various improvements and customization options

### DIFF
--- a/packages/flutter_link_previewer/lib/src/utils.dart
+++ b/packages/flutter_link_previewer/lib/src/utils.dart
@@ -348,4 +348,4 @@ const regexImageContentType = r'image\/*';
 
 /// Regex to find all links in the text.
 const regexLink =
-    r'((http|ftp|https):\/\/)?(([\p{L}\p{N}_-]+)(?:(?:\.([\p{L}\p{N}_-]*[\p{L}_][\p{L}\p{N}_-]*))+))([\p{L}\p{N}.,@?^=%&:/~+#-]*[\p{L}\p{N}@?^=%&/~+#-])?[^\.\s]';
+    r'(?<!href="|">)(https?:\/\/[^\s<]+)';


packages/flyer_chat_text_message/lib/src/flyer_chat_text_message.dart
```diff
@@ -123,6 +123,9 @@ class FlyerChatTextMessage extends StatelessWidget {
   /// The widget to display on top of the message.
   final Widget? topWidget;
 
+  /// Widget to display as text, overriding message text.
+  final Widget? customWidget;
```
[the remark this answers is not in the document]
Henry-Hiles: The thing is, I want all the features of the FlyerChatTextMessage, and the nice UI, but I need to be able to render HTML.

packages/flutter_chat_ui/lib/src/composer.dart
```diff
@@ -132,6 +132,13 @@ class Composer extends StatefulWidget {
   /// and hide the button. Defaults to `false`.
   final bool sendButtonHidden;
 
+  /// Whether to send messages on enter.
+  ///
+  /// If `true`, typing enter will send the message, and typing shift+enter
+  /// will type enter. If `false`, typing shift+enter will send the message.
+  /// Defaults to `false`.
+  final bool sendOnEnter;
```
demchenkoalex: this feels incomplete. right now `shift+enter` will also send a message and enter will make a new line. if you want to send on enter - how user makes a new line?

Henry-Hiles: This variable makes it so that the message sends on enter, and shift+enter only adds a new line.

packages/flyer_chat_text_message/lib/src/flyer_chat_text_message.dart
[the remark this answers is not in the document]
demchenkoalex: yep, just copy the code and use HTML :) I specifically created the builders pattern to avoid all the "customXXX" stuff that was everywhere in v1 - people would throw all sorts of things in there and then complain the package didn't work. Now, when someone needs something super custom, I prefer to give them full control and responsibility.

packages/flutter_chat_ui/lib/src/composer.dart
demchenkoalex: I saw the change to send on enter, but did not see the code for the new line itself - is it automatic?

Henry-Hiles: I used an XOR operator for it on line 222 of composer.dart, inside your existing logic for this.

packages/flyer_chat_text_message/lib/src/flyer_chat_text_message.dart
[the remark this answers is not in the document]
Henry-Hiles: Hmm, but thats a lot of code I need to copy, when I only want to change the display of one widget...
I guess I can just maintain a fork of this for my purposes.

packages/flutter_chat_ui/lib/src/composer.dart
Henry-Hiles: It definitely works, I'm using it in my app right now.

Henry-Hiles:
> Reopening since it's something I would need

The send on enter, only?

demchenkoalex:
Released in [v2.11.0](https://github.com/flyerhq/flutter_chat_ui/releases/tag/v2.11.0)

PR #893, opened 2026-09-02
Replace flutter imports with material_ui, cupertino_ui.

riyaz-raz:
- Replace flutter/material.dart imports with material_ui
- Replace flutter/cupertino.dart imports with cupertino_ui


timmywil:
Thanks for doing this! This is the last package I need migrated in order to migrate my app off legacy material.

demchenkoalex:
published. sorry I no longer have time to maintain this so I open GitHub rarely.

